PR #1, opened 2019-05-31
Input Manager for horizontal, vertical and breaking input.

m4rtint:
Input manager with Horizontal, vertical and car break input
Removed unnecessary add ons
Changed some settings in the Project Settings > Input.
.gitignore added - StyleCop.Cache

// <auto-generated/>
means that stylecop will ignore these files


m4rtint:
To use this manager, in your Update() Method, just call 
InputManger.IsBreaking(), InputManager.MainHorizontal()...etc.  to check if the user has pressed any buttons

mikeslgoh:
[Input Manager](https://trello.com/c/0P2b1iqw/17-input-manager)

PR #2, opened 2019-06-04
Basic Car Control works. Worth polishing later

wbeltran777:
Found a great tutorial for writing a car controller. 
It can definitely be polished a bit more, but I'm uploading in case anyone else needs a car controller for other tasks. We can definitely modify it more in the future.


m4rtint:
In the root folder, there is a .gitignore file,
Add the line
.DS_Store at the bottom

and then remove all the .DS_Store files and then recommit

Also, in the Assets folder, there is a .gitignore file @mikeslgoh  is that intentional?

Assets/Materials/Car.mat
```diff
@@ -0,0 +1,77 @@
+%YAML 1.1
```
m4rtint: How come there is car materials? i thought you just need the sprite?

m4rtint:
Don't forget to run the style cop plugin, I see you're missing out on quite a lot of coding standards. e.g. extra spaces, not using "this" connotations...etc

m4rtint:
[Car movement](https://trello.com/c/RueB2zj1/14-car-movement)

Assets/Script/CarController.cs
```diff
@@ -0,0 +1,75 @@
+﻿//-----------------------------------------------------------------------
+// <copyright file="InputManager.cs" company="GOTTY">
```
m4rtint: Don't forget to change this to CarController.cs

Assets/Materials/Car.mat
m4rtint: And how come there is a 3d object on the scene?

mikeslgoh:
@m4rtint that was set up by Unity when I installed the plugin. I've opened that ignore file before and I don't think there is anything written inside it. Maybe its in case we want to ignore files in our asset folder?

Assets/Materials/Car.mat
wbeltran777: Materials is useless, I'll remove it and recommit. I was using them before I got the sprites. 
The 3D object is a cube to have any kind of object in the background so we can see the car moving relative to something else. Should I remove it?

Assets/Script/CarController.cs
wbeltran777: Woops you right, I'll change that

Assets/Materials/Car.mat
m4rtint: Yes please. Including the materials which i'm guessing are also not needed. 

Next time you can make a 2d sprite object  🐥

m4rtint:
@mikeslgoh I assume the root .gitignore covers everything inside?

mikeslgoh:
Yeah the .gitignore used is the one in the root folder

Assets/Script/CarController.cs
```diff
@@ -0,0 +1,75 @@
...
+    {
+        rigidbody2D = GetComponent<Rigidbody2D>();
+    }
+
+
+    /// <summary>
+    /// Check for inputs from InputManager. 
+    /// Modify car acceleration accordingly. 
+    /// </summary>
+    void Update()
```
m4rtint: I think this should be FixedUpdate() ? @mikeslgoh @Gursaanj @spicyroboto what do you think?

Gursaanj: Anything with physics involved should have FixedUpdate(); for more accurate time based calculations. I think!
Since it is a car, we will probably add some physics to it eventually. so yeah maybe FixedUpdate(); could be the better choice

spicyroboto: Agree with Gursaanj!

wbeltran777:
> In the root folder, there is a .gitignore file,
> Add the line
> .DS_Store at the bottom
> 
> and then remove all the .DS_Store files and then recommit
> 
> Also, in the Assets folder, there is a .gitignore file @mikeslgoh is that intentional?

How do I access the gitignore file? Do you mean literally type a line at the bottom that says .DS_Store? What does this do?

m4rtint:
> > In the root folder, there is a .gitignore file,
> > Add the line
> > .DS_Store at the bottom
> > and then remove all the .DS_Store files and then recommit
> > Also, in the Assets folder, there is a .gitignore file @mikeslgoh is that intentional?
> 
> How do I access the gitignore file? Do you mean literally type a line at the bottom that says .DS_Store? What does this do?

Yes literally type a line at the bottom ".DS_Store".
It ignore the file .DS_Store from being committed.  so that it doesn't get committed and pushed. 
But because you already commited .DS_Store, you'll need to remove it.

PR #3, opened 2019-06-06
Added Simple NPC that drives forward and backward

Gursaanj:
Created a NPC abstract class for all future and current vehicle NPC to derive from, giving some fundament public methods as well as some properties that all NPC should have (RigidBody2D , Collider2D ...) and I created one prefab called simple car which is a vehicle NPC that (at a given angle from the vertical) will move back and forth for a given distance at a given speed. Ofcourse alot more newer NPC can be made now (with different functionality) but that will be added in over time with the consensus  of the group.  I plan to create another Abstract Class for just static objects within the coming days


Assets/Script/SimpleCar.cs
```diff
@@ -0,0 +1,90 @@
...
+    private float rotation;
+
+    /// <summary> 
+    /// At the beginning the RigidBody2D (and the collider technically)
+    /// is attached to the script. The rotation of the object is set 
+    /// and the distance from the origin of the simple car is calculated 
+    /// </summary>
+    private void Start()
+    {
+        this.rb = gameObject.GetComponent<Rigidbody2D>();
```
m4rtint: Can you place all the  gameObject.GetComponent stufff into Awake();

Assets/Script/SimpleCar.cs
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    /// <summary>
+    /// Set up the movement of the NPC to update per physical frame
+    /// When the NPC reaches the determined distance from its starting position
+    /// reverse the motion
+    /// </summary>
+    private void FixedUpdate()
+    {
+        this.Translate(this.speed, Time.deltaTime);
```
m4rtint: Can you place these into separate methods?

Assets/Script/SimpleCar.cs
```diff
@@ -0,0 +1,90 @@
...
+
+        if (this.Distance(this.transform.position.x, this.transform.position.y) - this.startDistance >= this.distance)
```
m4rtint: You can place this inside change Direction() method
`if (this.Distance(this.transform.position.x, this.transform.position.y) - this.startDistance >= this.distance) {
this.speed *= -1
}`

Assets/Scenes/SampleScene.unity
```diff
@@ -50,7 +50,6 @@ LightmapSettings:
     m_BounceScale: 1
```
m4rtint: Would you know how to uncommiit this?

m4rtint: This file i meant

Assets/Script/VehicleNPC.cs
```diff
@@ -0,0 +1,47 @@
...
+{
+    /// <summary>
+    /// Determines the distance of the NPC using the global x and y position via pythagoras 
+    /// </summary>
+    /// <param name="x">The x distance from origin</param>
+    /// <param name="y">The y distance from origin</param>
+    /// <returns>The direct distance from origin</returns>
+    public float Distance(float x, float y)
+    {
+        float z = Mathf.Sqrt(Mathf.Pow(x, 2) + Mathf.Pow(y, 2));
```
m4rtint: No need to store this in variable z, just return this

Assets/Script/SimpleCar.cs
m4rtint: btw, no need to write gameObject, can just call GetComponent

mikeslgoh:
[Simple NPC/obstacles](https://trello.com/c/lD4CyGtL/18-simple-npc-obstacles)

Assets/Scenes/SampleScene.unity
Gursaanj: hmm Im not sure, Ill look online and see if I can

If I cant someone can just help me do it today sometime during the meeting haha

mikeslgoh: what is this exactly? To uncommit, you can just remove whatever that file is and then commit that change to your branch. It should remove it from your commit @Gursaanj

m4rtint: @Gursaanj  a super simple way to do this is to 
1)checkout from develop
2)copy the SampleScene.unity file somewhere
3)checkout this branch/pullrequest
4)paste this unity scene back.

this would overwrite the file to be back to whatever develop was

Gursaanj: So just copy the SampleScene.unity from develop and then replace the one in my branch with that. Then recommit?

m4rtint: yeah that works

Assets/Script/SimpleCar.cs
Gursaanj: You mean take out Translate from FixedUpdate() ?

Assets/Script/SimpleCar.cs
Gursaanj: Do you mean that I should put  only the collider2D and the RigidBody2D in the void Awake and keep the rest in the void Start. 

Or do you mean I should just scrap Void Start and use void Awake only?

mikeslgoh: @Gursaanj I think he meant keep START still for things like .Rotate and this.Distance

Assets/Script/SimpleCar.cs
mikeslgoh: I would recommend this.Translate is one method and the calculation for your IF statement is another method since it isn't a simple comparison. @m4rtint is that right?

Assets/Script/SimpleCar.cs
Gursaanj: Okay Thought so. Will do  Thanks

Assets/Script/SimpleCar.cs
Gursaanj: I would agree with @mikeslgoh  but maybe I don't the most efficient way of doing it

m4rtint: @Gursaanj  Basically what @mikeslgoh said. 
I guess this one line isn't a huge deal. but for sure the if statement in another method, so we know what's going on in the if statement.

Gursaanj: For sure, I placed the If statement in the Change direction Method. So right now it is just the Translate() and the ChangeDirection() Methods in the FixedUpdate()

mikeslgoh: Should be fine then. Yeah the one line shouldn't be an issue since I think its explained in your summary tag

Gursaanj: Sounds good. Thanks

Assets/Script/SimpleCar.cs
```diff
@@ -0,0 +1,97 @@
...
+    [Range(0, 360)]
+    private float rotation;
+
+    /// <summary>
+    /// At the beginning the RigidBody2D (and the collider technically)
+    /// is attached to the script.
+    /// </summary>
+    private void Awake()
+    {
+        this.rb = gameObject.GetComponent<Rigidbody2D>();
```
m4rtint: You can remove the gameObject variable. 


Also this.rb.gravityScale could be in Start()

Assets/Script/SimpleCar.cs
```diff
@@ -0,0 +1,97 @@
...
+
+        this.ChangeDirection();
+    }
+
+    /// <summary>
+    /// Reverses the motion of the NPC at given condition
+    /// </summary>
+    private void ChangeDirection()
+    {
+        if (this.Distance(transform.position.x, transform.position.y) - this.startDistance >= this.distance)
```
m4rtint: I'm surprised stylecop don't complain over transform.position.x

thought they wanted this. 
 oh well

PR #4, opened 2019-06-06
Simple indicator for car

mikeslgoh:
Created a simple indicator that uses an arrow sprite to direct the player to a target (parking spot)


Assets/Script/ParkingIndicator.cs
```diff
@@ -0,0 +1,66 @@
...
+
+    /// <summary>
+    /// Sets the children active.
+    /// </summary>
+    /// <param name="value"> Set children to either true or false </param>
+    private void SetChildrenActive(bool value)
+    {
+        foreach (Transform child in this.transform)
+        {
+            child.gameObject.SetActive(value);
```
m4rtint: What is the child of this gameObject? 

fun fact - in this case for (int i = 0 ; i < size; i++){ }
this is more efficient for this case, since it's not creating an extra temp variable "child", so unless you use child more than once, it's less efficient using foreach. 

No need to change this, just something Ryan Oh told me about.

mikeslgoh: The child is the arrow sprite. In the tutorials I found, they did not apply the script onto the arrow sprite itself but on the parent of the sprite. Also, that might something I make a change to when I change this implementation since it is rather simple still.

m4rtint: okay that makes sense. Separating the logic/script from the UI/Sprite

mikeslgoh:
[Indicator](https://trello.com/c/gpuPHt7F/15-indicator)

PR #5, opened 2019-06-07
Added simple bounds to check if car is in parking spot

spicyroboto:


spicyroboto:
[Detection between car and parking spot](https://trello.com/c/jQtNI4K7/13-detection-between-car-and-parking-spot)

spicyroboto:
[Detection between car and parking spot](https://trello.com/c/jQtNI4K7/13-detection-between-car-and-parking-spot)

Assets/Script/ParkingDetection.cs
```diff
@@ -0,0 +1,69 @@
...
+        Bounds parkingSpotBounds = parkingSpot.bounds;
+
+        Vector2 carCenter = carBounds.center;
+        Vector2 carEdges = carBounds.extents;
+        Vector2[] carVerticies = new Vector2[4];
+
+        carVerticies[0] = new Vector2(carCenter.x + carEdges.x, carCenter.y + carEdges.y);
+        carVerticies[1] = new Vector2(carCenter.x - carEdges.x, carCenter.y + carEdges.y);
+        carVerticies[2] = new Vector2(carCenter.x + carEdges.x, carCenter.y - carEdges.y);
+        carVerticies[2] = new Vector2(carCenter.x - carEdges.x, carCenter.y - carEdges.y);
```
m4rtint: index should be 3

m4rtint:
if(topHeaderBoxCollider.bounds.Intersects(currentHeader.boxCollider.bounds))

This actually checks if one collider is inside another.


Another thing is that, did you ever try using 
void OnTriggerStay(){}

Assets/Script/ParkingDetection.cs
spicyroboto: Fixed

spicyroboto:
I'll look into #2. Hadn't seen that approach before. Chose to do it this way since it's the solution I was taught while doing interview prep for video games. 

For #3, I've seen a lot of people having issues with weird OnTriggerStay behaviour. Is there a particular way you'd want to use it?

m4rtint:
> I'll look into 2. Hadn't seen that approach before. Chose to do it this way since it's the solution I was taught while doing interview prep for video games.
> 
> For 3, I've seen a lot of people having issues with weird OnTriggerStay behaviour. Is there a particular way you'd want to use it?

Interesting, how come that's the solution for interview prep for video games?

and for 3, I'm not surprised if people think OnTriggerStay Behaivour is janky. I'll trust that it is weird enough that your method more accurate and efficient.

m4rtint:
Also merge conflict.

Assets/Script/ParkingDetection.cs
```diff
@@ -0,0 +1,69 @@
...
+/// <summary>
+/// Handles bounds checking between the user's car and a parking spot
+/// </summary>
+public class ParkingDetection : MonoBehaviour
+{
+    /// <summary>
+    /// The player's car
+    /// </summary>
+    [SerializeField]
+    private BoxCollider2D car;
```
mikeslgoh: shouldn't we start using m_ for the fields?

m4rtint: Keep in mind, if my stylecop changes aren't merged in, stylecop will complain about m_

spicyroboto: Yeah, I originally put the m_, but it complained at me because Stylecop hadn't been changed yet.

m4rtint:
@spicyroboto the new stylecop settings has been merged.

Assets/Script/ParkingDetection.cs
```diff
@@ -0,0 +1,69 @@
...
+    /// <summary>
+    /// The target (aka the parking spot) the player is navigating towards
+    /// </summary>
+    [SerializeField]
+    private BoxCollider2D parkingSpot;
+
+    /// <summary>
+    /// Checks in each update loop whether the car has successfully parked in a parking spot
+    /// </summary>
+    private void Update()
```
Gursaanj: Should all our scripts be using FixedUpdate() instead of Update() ? Technically this shouldn't matter since its not really a physics thing, but would it be better for consistency sake to keep all our Update methods to be FixedUpdate? Just Wondering

spicyroboto: Good question. I don’t have enough experience with it to have an opinion. @mikeslgoh and @m4rtint ?

m4rtint: Yeah I think this should be FixedUpdate() 
since it's related to gameplay.

Assets/Script/ParkingDetection.cs
```diff
@@ -0,0 +1,69 @@
...
+    private bool IsParked(BoxCollider2D car, BoxCollider2D parkingSpot)
+    {
+        Bounds carBounds = car.bounds;
+        Bounds parkingSpotBounds = parkingSpot.bounds;
+
+        Vector2 carCenter = carBounds.center;
+        Vector2 carEdges = carBounds.extents;
+        Vector2[] carVerticies = new Vector2[4];
+
+        carVerticies[0] = new Vector2(carCenter.x + carEdges.x, carCenter.y + carEdges.y);
```
m4rtint: this is a bit of a nitpick, which u can probably ignore. There is a property in bound, which is bound.max.
which is essentially what you wrote. But I think it's better to stick with this for readability.

Assets/Script/ParkingDetection.cs
mikeslgoh: Yup agree ^

Assets/Script/ParkingDetection.cs
mikeslgoh: Yeah I realised but that's why I left a comment here coz I just merged it

m4rtint: @mikeslgoh ahh u approved it, not yet merged 🐔

PR #8, opened 2019-06-09
Basic Car Controller

wbeltran777:
Touched it up more than last time, feels pretty good! 

Revised my style cop comments. Ignored the "this." and the headers for each field as we discussed we'd be ignoring those.


wbeltran777:
[Car movement](https://trello.com/c/RueB2zj1/14-car-movement)

m4rtint:
You should merge the develop branch into your branch.

and fix conflicts

Assets/Script/CameraFollower.cs
```diff
@@ -0,0 +1,28 @@
...
+
+/// <summary>
+/// Allows Main Camera to follow car as the player drives around. 
+/// </summary>
+public class CameraFollower : MonoBehaviour
+{
+    /// <summary>
+    /// Target the camera must follow. Player in this case
+    /// </summary>
+    [SerializeField] private Transform target;
```
m4rtint: change this to m_target

Assets/Script/CarController.cs
```diff
@@ -0,0 +1,91 @@
...
+        m_currentTorqueForce = 0;
+        m_rigidbody2D = GetComponent<Rigidbody2D>();
+    }
+
+
+    /// <summary>
+    /// Check for inputs from InputManager. 
+    /// Modify car acceleration accordingly. 
+    /// </summary>
+    void Update()
```
m4rtint: FixedUpdate()

Assets/Script/CameraFollower.cs
```diff
@@ -0,0 +1,27 @@
...
+{
+    /// <summary>
+    /// Target the camera must follow. Player in this case
+    /// </summary>
+    [SerializeField] private Transform m_target;
+
+    /// <summary>
+    /// Adjust position of camera to always be on the players transform coordiantes
+    /// </summary>
+    private void Update()
```
mikeslgoh: Didn't we say we should be using FixedUpdate?

Gursaanj: If the CarController is using FixedUpdate() too I don't think there should be any reason that the Camera should be doing it too. 
But then again not sure it will make too much of a difference. Changing to FixedUpdate() would just be for more consistency I guess

PR #9, opened 2019-06-12
Feature/splitscreen

mikeslgoh:
Created the code for splitscreen. The SplitScreenManager needs to be passed how many players are playing and it'll split the screen according. 

Would like some feedback on my approach and whether there are things I should be aware of/change

NOTE: Will styleCop as soon as I am able to


mikeslgoh:
[SplitScreen](https://trello.com/c/3E3SG5MU/25-splitscreen)

Assets/Script/ParkingIndicator.cs
```diff
@@ -63,4 +63,9 @@ private void CalculateIndicatorRotation(Vector2 dir)
         float angle = Mathf.Atan2(dir.y, dir.x) * Mathf.Rad2Deg;
         transform.rotation = Quaternion.AngleAxis(angle, Vector3.forward);
     }
+
```
m4rtint: This would become a property I assume

Assets/Script/ScreenAttributes.cs
```diff
@@ -0,0 +1,13 @@
+﻿using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+using System;
+
+[Serializable]
+public class ScreenAttributes
```
m4rtint: How come this is Serializable? 

And would a struct be better in this case?

Assets/Script/SplitScreenManager.cs
```diff
@@ -0,0 +1,101 @@
...
+        newPlayer.name = "Player" + (playerNum+1);
+
+        GameObject car = Instantiate(m_carTemplate);
+        setupPlayerCamera(car, newPlayer, playerNum);
+
+        car.transform.parent = newPlayer.transform;
+        car.transform.localPosition = new Vector3(0, 0, 0);
+        car.transform.localRotation = m_spawnPoints[playerNum].rotation;
+
+        GameObject indicator = Instantiate(m_indicatorTemplate);
```
m4rtint: I think the indicator could be attached to the car prefab
So this manager shouldn't have to handle all this stuff.

And also maybe the car should have its own class to handle its own setup. Just hand it the spawn point and let it do everything else.

Assets/Script/SplitScreenManager.cs
```diff
@@ -0,0 +1,101 @@
...
+        foreach (Transform child in spawnPointContainer.transform)
+        {
+            m_spawnPoints.Add(child);
+        }
+
+        m_screenAttrFile = Path.Combine(Application.streamingAssetsPath, m_screenAttrFile);
+
+        if (m_numOfPlayers > 4 || m_numOfPlayers < 1)
+        {
+            Debug.LogError("Number of players not between 1 and 4");
```
m4rtint: Check Unity to see if there is some assert function that doesn't allow running the game unless it's between 1 and 4

Assets/Script/SplitScreenManager.cs
```diff
@@ -0,0 +1,101 @@
...
+
+        string dataAsJson = File.ReadAllText(m_screenAttrFile);
+        NumberOfPlayers configNumPlayers = JsonUtility.FromJson<NumberOfPlayers>(dataAsJson);
+
+      if (m_numOfPlayers == 3 || m_numOfPlayers == 4)
+        {
+            GameObject playerCamera = setupCamera(car);
+            ScreenAttributes screen = configNumPlayers.fourPlayer[playerNum];
+            playerCamera.GetComponent<Camera>().rect = new Rect(screen.x, screen.y, screen.width, screen.height);
+            playerCamera.transform.parent = newPlayer.transform;
```
m4rtint: So is the camera pegged to the car?

Assets/StreamingAssets/screenAttributes.json
```diff
@@ -0,0 +1,41 @@
+{
+  "twoPlayer": [
+  {
+      "width": 1,
+      "height": 0.5,
+      "x": 0,
+      "y": 0.5
```
m4rtint: Are 2 players top and bottom screen?  or is this side by side?

Assets/Script/SplitScreenManager.cs
Gursaanj: Yeah that idea feels a lot cleaner

Assets/Script/SplitScreenManager.cs
Gursaanj: https://docs.unity3d.com/ScriptReference/Assertions.Assert.html

Like this?

Assets/StreamingAssets/screenAttributes.json
Gursaanj: I feel that if it is two players, it should be side by side rather that top and bottom.

Assets/Script/CameraFollower.cs
```diff
@@ -17,11 +17,23 @@ public class CameraFollower : MonoBehaviour
     /// </summary>
     [SerializeField] private Transform m_target;
 
+    [SerializeField] private float m_distance;
+
     /// <summary>
     /// Adjust position of camera to always be on the players transform coordiantes
     /// </summary>
     private void Update()
```
Gursaanj: I know this isnt the final version, and if everyone accepts Martins WolrdManager class we dont have to worry about updates, but I guess this should FixedUpdate() too

Assets/Script/ParkingIndicator.cs
mikeslgoh: do you mean the angle variable?

Assets/Script/ScreenAttributes.cs
mikeslgoh: Yeah was searching stuff up on Google and forgot C# can do structs. Will make that change and test

Assets/Script/SplitScreenManager.cs
mikeslgoh: So the camera is under a Player parent object and is then linked to the car

Assets/StreamingAssets/screenAttributes.json
mikeslgoh: Yeah I didn't know what would work better

Assets/Script/ParkingIndicator.cs
m4rtint: Sorry i dind't click on the correct line, 
setTarget() method.

Assets/Script/SplitScreenManager.cs
m4rtint: yeah like that

Assets/Script/SplitScreenManager.cs
```diff
@@ -0,0 +1,101 @@
...
+        indicator.GetComponent<ParkingIndicator>().setTarget(m_initialTarget.transform);
+        indicator.transform.parent = car.transform;
+        indicator.transform.localPosition = new Vector3(0, 0, 0);
+    }
+
+    private void setupPlayerCamera(GameObject car, GameObject newPlayer, int playerNum)
+    {
+
+        string dataAsJson = File.ReadAllText(m_screenAttrFile);
+        NumberOfPlayers configNumPlayers = JsonUtility.FromJson<NumberOfPlayers>(dataAsJson);
```
Gursaanj: Not really a comment, just a question 

How does this work? Have never really imported files to be read by Unity before

Assets/Script/ParkingIndicator.cs
mikeslgoh: Yeah I can make that change

Assets/Script/CameraFollower.cs
mikeslgoh: Yeah haha. This wasn't my script but I'll make the change

Assets/Script/SplitScreenManager.cs
mikeslgoh: Basically, the information is read from a JSON file I stored in StreamingAssets and is then matched to the class NumberOfPlayers. This would work even when we do a build of the project

Assets/StreamingAssets/screenAttributes.json
m4rtint: Do side by side, top and bottom u see too much. Depends on the map really. but i rather stick with side by side if we want to stick with landscape maps

Assets/Script/SplitScreenManager.cs
m4rtint: If it's linked to the car, would is still need the camera follower?

Assets/Script/ScreenAttributes.cs
mikeslgoh: So turns out this does not be serializable or else when I read it from the JSON file, it's not able to create it but I have changed this to a struct

Assets/Script/SplitScreenManager.cs
mikeslgoh: The camera follower makes the camera stable and not rotate around the car. If you try opening my splitscreen and playing with a car, you can see for yourself as well.

m4rtint: oh that makes sense. okay sure thing.

Assets/Script/CarController.cs
```diff
@@ -40,6 +40,16 @@ private void Awake()
         m_rigidbody2D = GetComponent<Rigidbody2D>();
     }
 
+    public static GameObject createCar(Transform spawnPoint, Transform target, GameObject carTemplate)
```
m4rtint: this shouldn't be static anymore. 

set the localPosition here. Vector3.zero works

Assets/Script/CarController.cs
```diff
@@ -40,6 +40,16 @@ private void Awake()
...
+    {
+        GameObject car = Instantiate(carTemplate);
+        car.transform.localRotation = spawnPoint.rotation;
+        GameObject indicator = car.transform.Find("ParkingIndicator").gameObject;
```
m4rtint: I think the Car Controller should have access to the indicator
[SerializedField] ParkingIndicator indicator

The indicator should be part of the car prefab.

m4rtint: FYI - the Find() method is super bad in general. 

Its run time is O(n), it goes through the whole list trying to find your string.

Assets/Script/SplitScreenManager.cs
```diff
@@ -0,0 +1,99 @@
...
+        {
+            createPlayer(i);
+        }
+    }
+
+    private void createPlayer(int playerNum)
+    {
+        GameObject newPlayer = new GameObject();
+        newPlayer.transform.position = m_spawnPoints[playerNum].position; 
+        newPlayer.name = "Player" + (playerNum+1);
```
m4rtint: Put the string as a readonly on top

Assets/Script/SplitScreenManager.cs
```diff
@@ -0,0 +1,99 @@
...
+
+    private void setupPlayerCamera(GameObject car, GameObject newPlayer, int playerNum)
+    {
+
+        string dataAsJson = File.ReadAllText(m_screenAttrFile);
+        NumberOfPlayers configNumPlayers = JsonUtility.FromJson<NumberOfPlayers>(dataAsJson);
+
+      if (m_numOfPlayers == 3 || m_numOfPlayers == 4)
+        {
+            GameObject playerCamera = setupCamera(car);
```
m4rtint: place this outside the if statement on top. No need to repeat so many times

Assets/Script/SplitScreenManager.cs
```diff
@@ -0,0 +1,99 @@
...
+            ScreenAttributes screen = configNumPlayers.twoPlayer[playerNum];
+            playerCamera.GetComponent<Camera>().rect = new Rect(screen.x, screen.y, screen.width, screen.height);
+            playerCamera.transform.parent = newPlayer.transform;
+        }
+        else
+        {
+            GameObject playerCamera = setupCamera(car);
+            playerCamera.transform.parent = newPlayer.transform;
```
m4rtint: Place this line outside the if statement on the bottom, it's going to be hit no matter what.

Assets/Script/CameraFollower.cs
```diff
@@ -17,11 +17,23 @@ public class CameraFollower : MonoBehaviour
...
     {
-        this.transform.position = new Vector3(m_target.position.x, m_target.position.y, -10f);
+        transform.position = new Vector3(m_target.position.x, m_target.position.y, m_distance);
+    }
+
+    public void setTarget(Transform target)
```
m4rtint: setTarget and setDistance should also be properties

Assets/Script/SplitScreenManager.cs
```diff
@@ -0,0 +1,92 @@
...
+        setupPlayerCamera(playerCar, newPlayer, playerNum);
+    }
+
+    private void setupPlayerCamera(GameObject car, GameObject newPlayer, int playerNum)
+    {
+
+        string dataAsJson = File.ReadAllText(m_screenAttrFile);
+        NumberOfPlayers configNumPlayers = JsonUtility.FromJson<NumberOfPlayers>(dataAsJson);
+        GameObject playerCamera = setupCamera(car);
+
```
m4rtint: You can also do
`ScreenAttributes screen = null`

if (numplayer ==3 || 4)
`screen = configNum.fourPlayer`
else if (numPlayer == 2) {
`screen = configNum.TwoPlayer`
}
```
if (screen != null){
playerCamera.GetComponent<Camera>().rect = new Rect(screen stuff);
}
```

mikeslgoh: ah that works too

mikeslgoh: For some reason, I can't declare my ScreenAttributes struct null so i'm not sure whether i should use a bool flag instead?

m4rtint: You can leave it then, if it doesn't work that way. 

There is also the option to make it a null object, but that's too much just for a tiny bit of efficiency.

m4rtint: ScreenAttributes? screen = null; 
declare it like that. not too bad, didn't know C# can do this

mikeslgoh: yeah I just tried and it complains about accessing the screen variables

Assets/Script/CameraFollower.cs
```diff
@@ -16,12 +16,46 @@ public class CameraFollower : MonoBehaviour
...
```
m4rtint: Oh you can remove this stuff since you put in the properties

mikeslgoh: I'm setting the properties but I need the transform.position to make sure the camera actually follows the player's movements. I tried removing it and it doesn't work without it

m4rtint: do you do like CameraFollower.Target = whateverTarget? 

the property should handle it.

mikeslgoh: That sets the target but I need onUpdate to update the camera position to move when the car moves because my splitscreen manager does not have an OnUpdate.

m4rtint: I'm only talking about 
public void setTarget(Transform target) method
public void setDistance(float dist)

mikeslgoh: oh my bad sorry. The main update function was in focus so I thought it was referring to that. Will remove the other set functions.

Assets/Script/Managers/WorldManager.cs
```diff
@@ -58,6 +71,57 @@ private void Awake()
     /// </summary>
     private void FixedUpdate()
     {
+        if (readyToPlay && !gameCreated)
+        {
+            SplitScreenSetup();
```
m4rtint: Why not put this in Start()?

Assets/Script/Managers/WorldManager.cs
```diff
@@ -28,6 +29,18 @@ public class WorldManager : MonoBehaviour
     /// </summary>s
     public static event WorldDelegate OnFixedUpdate;
 
+    private readonly string namePrefix = "Player";
+    private List<Transform> m_spawnPoints = new List<Transform>();
+    [SerializeField] private SplitScreenManager m_splitScreenManager;
+    [SerializeField] private int m_numOfPlayers;
+    [SerializeField] private GameObject m_carTemplate;
+    [SerializeField] private Transform m_initialTarget;
+    [SerializeField] private GameObject m_spawnPointContainer;
```
m4rtint: I'm assuming this is for the cars? Can you rename is to carSpawnPoint(s)?

Assets/Script/Managers/WorldManager.cs
mikeslgoh: Yeah didn't think of creating a start. Will move it

Assets/Script/Managers/WorldManager.cs
```diff
@@ -53,11 +66,66 @@ private void Awake()
         DontDestroyOnLoad(gameObject);
     }
 
+    private void Start()
+    {
+        if (readyToPlay && !gameCreated)
```
m4rtint: you shouldn't need readyToPlay and gameCreated boolean anymore

Assets/Script/Managers/WorldManager.cs
```diff
@@ -53,11 +66,66 @@ private void Awake()
...
+        GameObject playerCar = CreateCar(m_carSpawnPoints[playerNum]);
+        playerCar.transform.parent = newPlayer.transform;
+
+        m_splitScreenManager.SetupPlayerCamera(playerCar, newPlayer, playerNum, m_numOfPlayers, m_cameraDistance);
+    }
+
+    private GameObject CreateCar(Transform spawnPoint)
+    {
+        GameObject car = Instantiate(m_carTemplate, spawnPoint);
+        car.transform.GetChild(1).gameObject.GetComponent<ParkingIndicator>().Target = m_initialTarget;
```
m4rtint: Hardcoding the child is bad. 

Pass the m_initialTarget to the CarController to handle. 

And then CarController should handle the Parking Indicator.

Assets/Script/Managers/WorldManager.cs
mikeslgoh: Right ok. I originally set that up when it was in update so that the SplitScreenManager won't be called everytime

m4rtint: lol yup, i figured

mikeslgoh: We might need a flag in the future once we start having the level selection and so on but that's for a future implementation

m4rtint: I don't think we'll need to make the World Manager at the main menu. 
Since we're using a seperate static class to store all the data. 
So the World Manager should only be in the game scene.

Assets/Script/Managers/SplitScreenManager.cs
```diff
@@ -0,0 +1,54 @@
...
+using UnityEngine;
+
+/// <summary>
+/// Manages the splitscreen setup for the game
+/// </summary>
+public class SplitScreenManager : MonoBehaviour
+{
+    private string m_screenAttrFile = "screenAttributes.json";
+
+    public void SetupPlayerCamera(GameObject car, GameObject newPlayer, int playerNum, int numberOfPlayers, float cameraDistance)
```
wbeltran777: What's the purpose of having a car game object and a newplayer game object? Is the car getting assigned to the new player? Would this function be called for each player then or just once?

Assets/Script/NumberOfPlayers.cs
```diff
@@ -6,5 +6,6 @@
 public class NumberOfPlayers
```
wbeltran777: Would this work as a struct instead, since there's so few fields?

Assets/Script/Managers/SplitScreenManager.cs
mikeslgoh: The new player object is the parent. So this was using the prefab style that you had before with a parent object holding the Car object. This helps me to keep each player's camera under their own Player object as well. Hope that makes sense?

Assets/Script/NumberOfPlayers.cs
mikeslgoh: I did it as a class because the data stored aren't too closely related to one another whereas I usually associated structs to contain data is closely related to one another such as the ScreenAttributes. But not sure, @m4rtint @spicyroboto @Gursaanj thoughts?

m4rtint: I think I read somewhere that structs are short term, classes are long term.

PR #10, opened 2019-06-14
Refactor/delegate

m4rtint:


mikeslgoh:
do we need to StyleCop for Awake?

m4rtint:
> do we need to StyleCop for Awake?

What do you mean?

Assets/Script/CameraFollower.cs
```diff
@@ -17,11 +17,21 @@ public class CameraFollower : MonoBehaviour
     /// </summary>
     [SerializeField] private Transform m_target;
 
+    private void Awake()
```
Gursaanj: Yeah, shouldn't Private void Awake() and Private void OnDestroy()  have the summary Descriptions

/// <summary>
///  Whatever
/// </summary>

m4rtint: It's a private method, it's supposed to be optional if you want to describe it or not  😄

mikeslgoh: perfect that's good to know

Gursaanj: Oh did not know that haha

PR #12, opened 2019-06-15
Feature/states

m4rtint:
MenuManager.cs -> Placeholder stuff
SplitScreen.unity -> Changed and added to the editor scenes settings

Refactored WorldManager
Removed -> screenAttributes.json, Added all the settings through NumberOfPlayers, Style cop wanted the Properties/Fields in Number of properties

StateManager has pause play function


mikeslgoh:
Did you delete the Spawnpoint prefab? Might need to delete it from the scene then if you did.

image

m4rtint:
> Did you delete the Spawnpoint prefab? Might need to delete it from the scene then if you did.
> 
> image

i deleted the prefab, but it's still used in the scene

mikeslgoh:
> > Did you delete the Spawnpoint prefab? Might need to delete it from the scene then if you did.
> > image
> 
> i deleted the prefab, but it's still used in the scene

ah ok. Gotcha. I like the way everything is set up now. Probably I would suggest placing all the managers under a parent MANAGER object. Just to keep everything together and clean?

m4rtint:
Yeah thats fine

Assets/Script/ScreenAttributes.cs
```diff
@@ -11,23 +11,43 @@
 /// </summary>
 public struct ScreenAttributes
 {
+    /// <summary>
+    /// The Distance of camera to the car
+    /// </summary>
+    public static float CameraDistance = -6;
```
mikeslgoh: Should we have a way to play with the CameraDistance through the inspector or is -6 from your findings the best distance for all players?

m4rtint: I have thought about that. But right now you have to stop the build, edit and run. 

So even if it's in the inspector it takes that many steps. No different from changing it through code. Which is why I left it in here. There is a chance that we need different distance depending on player size

m4rtint: This is more of a.. It is what it is kind of answer. Not like the best or worst way

mikeslgoh: Yeah I was thinking that the camera distance might vary depending on the number of players but we'll keep that in mind when playtesting.

m4rtint: Assets/Script/CameraFollower.cs

I made the m_distance [serialized] so that you can change it on run time.

m4rtint:
[WorldManager](https://trello.com/c/V1gbYaOG/31-worldmanager)

m4rtint:
[State Manager](https://trello.com/c/nTbuKU7G/35-state-manager)

Assets/MenuManager.cs
```diff
@@ -0,0 +1,52 @@
...
+using UnityEngine.SceneManagement;
+using UnityEngine.UI;
+
+/// <summary>
+/// Manages all UI elements on menu
+/// </summary>
+public class MenuManager : MonoBehaviour
+{
+    [SerializeField]
+    private Button[] m_PlayersButton;
```
wbeltran777: Just curious about the initialization of this array. Because it's Serializable, are the size and the buttons setup in the inspector?

Assets/Script/Managers/StateManager.cs
```diff
@@ -0,0 +1,169 @@
...
+        {
+            Destroy(gameObject);
+        }
+
+        DontDestroyOnLoad(gameObject);
+    }
+
+    private void PauseOnSceneRigidBody()
+    {
+        m_listOfRigidBodies = FindObjectsOfType<Rigidbody2D>();
```
wbeltran777: Wouldn't this operation take too long or is the number of rigidbodys that we have in one scene too small to really care?

Assets/Script/Managers/StateManager.cs
```diff
@@ -0,0 +1,169 @@
...
+            OnPause();
+        }
+    }
+
+    /// <summary>
+    /// Plays the game.
+    /// </summary>
+    public void PlayGame()
+    {
+        m_state = State.Playing;
```
wbeltran777: Would it be fair to create the instance of the Score Manager in this method? How encapsulated do we want our code?

m4rtint: pretty modular, if you want to create an instance of something, you should do it in World Manager. Since that is where everything in the game starts

Assets/Script/Managers/StateManager.cs
m4rtint: There were 2 options I could think of
Tag all the cars, obstacles with a new tag and FindObjectsWithTag()

Or this one.  I was leaning towards Tag everything that is movable with a tag, but I was just afraid we might need the tag system for something else, so I opted with the system which is a bit more flexible? but you are right the current operation is a bit ehh, could be better.

Assets/MenuManager.cs
m4rtint: yes, depends on how many items you drag in

Assets/Script/Managers/StateManager.cs
wbeltran777: Would we be able to create a list of dynamic rigidbodys right before the game starts? It could just be a loading screen operation and then everything after that should be smoother?

m4rtint: Good point. I'll think about what could be more efficient. 

Only issue would be if we were to add more rigid bodies to the scene on run time. 

But for now I'll change it to fit your idea. Thanks. =)

Assets/Script/Managers/StateManager.cs
```diff
@@ -0,0 +1,183 @@
...
+    }
+
+    private void PauseOnSceneRigidBody()
+    {
+        foreach (Rigidbody2D rb in m_listOfRigidBodies)
+        {
+            if (rb.bodyType == RigidbodyType2D.Dynamic)
+            {
+                m_listOfVelocity.Add(rb, rb.velocity);
+                rb.constraints = RigidbodyConstraints2D.FreezeAll;
```
Gursaanj: Also just found out the velocity/Movement of the A* NPC/objects is determined by the Pathfinding AI script not the Rigidbody. so freezing their Rigidbodies may not stop them?? Im not sure actually

m4rtint: u can test it out, by freezing it while it's moving

Assets/Script/Managers/WorldManager.cs
```diff
@@ -122,8 +95,32 @@ private void CreatePlayer(int playerNum)
...
 
         return car;
     }
+
+    /// <summary>
+    /// Setup the player count incase the GameData static class is not set
+    /// </summary>
+    private void SetupPlayerCount()
+    {
+        m_numberOfPlayers = (GameData.PlayerCount > 0) ? GameData.PlayerCount : m_numberOfPlayers;
```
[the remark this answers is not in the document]
m4rtint: if (GameData.playercount > 0)
m_numberOfPlayers = GameData.playercount
else
m_numberOfPlayers = m_numberOfPlayers

Assets/Script/Managers/StateManager.cs
```diff
@@ -0,0 +1,183 @@
...
+        }
+        else
+        {
+            Destroy(gameObject);
+        }
+
+        DontDestroyOnLoad(gameObject);
+    }
+
+    private void PauseOnSceneRigidBody()
```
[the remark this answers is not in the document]
m4rtint: if there's animation for pause panel, end game panel, it will not work

Assets/Script/Managers/StateManager.cs
```diff
@@ -0,0 +1,183 @@
...
+    /// <summary>
+    /// Gets list of rigid bodies on the scene
+    /// </summary>
+    public Rigidbody2D[] ListOfRigidBodies
+    {
+        get
+        {
+            if (m_listOfRigidBodies == null)
+            {
+                m_listOfRigidBodies = FindObjectsOfType<Rigidbody2D>();
```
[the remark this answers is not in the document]
m4rtint: Awake() for this script is run in mainmenu. 

so this one method is called only once in the game scene when it's needed

Assets/Script/Managers/StateManager.cs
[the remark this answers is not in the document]
Gursaanj: Fair point. sounds good

Assets/Script/Managers/StateManager.cs
Gursaanj: Okay edited my script to let the RigidBody2D freeze the movement. We good

PR #14, opened 2019-06-20
Feature/parking gen

spicyroboto:
DO NOT MERGE


Assets/Script/Managers/TargetManager.cs
```diff
@@ -0,0 +1,98 @@
...
+    public GameObject CurrentTarget { get; set; }
+    public BoxCollider2D Car { get; set; }
+
+    public void TargetManagerInitialization()
+    {
+        FindAllChildren();
+        ActivateRandomParkingSpot();
+    }
+
+    internal void ActivateRandomParkingSpot()
```
m4rtint: How come these are internal?

m4rtint:
Next time when making a pull request, make sure you set the base as Develop.

Assets/Script/Managers/TargetManager.cs
```diff
@@ -0,0 +1,98 @@
...
+        }
+    }
+
+    /// <summary>
+    /// Checks to see that all edges of the car are within the bounds of the parking spot 
+    /// </summary>
+    /// <param name="car"> The user's car </param>
+    /// <param name="parkingSpot"> The target parking spot</param>
+    /// <returns>True if car is in parking spot, false otherwise</returns>
+    private bool IsParked(BoxCollider2D car, BoxCollider2D parkingSpot)
```
m4rtint: So as far as i understand, we potentially have 2 ways of doing this.
1) Leave this here, grab the 4 cars from somewhere, and then loop through all 4 cars with this method. 
So sort of like
`foreach car inside ListOfCars
{ IsParked( car, this.parkingSpot) }`

Other option
2)Just like how it is originally, IsParked() is inside a separate script, attached to each parking spot
Inside OnTriggerStay(
So sort of like
`OnTriggerStay() {
if (gameObject.tag == "car") 
{
IsParked(car, gameObject);
}
}`

@wbeltran777 @mikeslgoh @Gursaanj

Gursaanj: The second method seems better to me since we will only need to call the method IsParked() when 
 a car has reached the parking spot.

wbeltran777: I agree with Gursaanj. The first method would have to continuously go through a for loop at every frame, no?

Assets/Script/Managers/WorldManager.cs
```diff
@@ -98,7 +102,7 @@ private GameObject CreateCar(Transform spawnPoint)
 
         // TODO this needs to be removed after parking spot randomizer is created
         car.GetComponent<CarController>().SetIndicatorTarget(m_initialTarget);
-
+        m_targetManager.Car = car.GetComponent<BoxCollider2D>();
```
wbeltran777: Car is a public field in Target Manager with getters and setters, but they're not used for this operation. Does C# take care of implicitly using the getters and setters based on how they're defined in TargetManager?

m4rtint: This is already using a setter. 
var car = m_targetManager.Car <--- this would be using the getter
m_targetManager.Car  = someCar <----this would be using a setter

wbeltran777: N e a t

m4rtint:
To get data from targetManager -> WorldManager

TargetManager
delegate onSpotReassigned(:Vector3)

WorldManager
onSpotReassigned += reassignCarTarget

void reassignCarTarget(Vector3) { for each car set indicator with Vector3) }

spicyroboto:
Awesome, thanks for your feedback guys!

mikeslgoh:
[Random generating the parking spot](https://trello.com/c/RaoUnJAh/19-random-generating-the-parking-spot)

Assets/Script/Managers/TargetManager.cs
```diff
@@ -0,0 +1,40 @@
...
+        }
+    }
+
+    private void FindAllChildren()
+    {
+        foreach (Transform child in transform)
+        {
+            m_targets.Add(child.gameObject);
+            child.gameObject.AddComponent<ParkingDetection>();
+            child.gameObject.GetComponent<ParkingDetection>().TargetManager = this;
```
m4rtint: To decouple this more, you can create a delegate in the ParkingDetection script

Assets/Script/Managers/WorldManager.cs
```diff
@@ -95,10 +99,10 @@ private void CreatePlayer(int playerNumber)
     private GameObject CreateCar(Transform spawnPoint)
     {
         GameObject car = Instantiate(m_carTemplate, spawnPoint);
+        car.tag = "playerCar";
```
m4rtint: Unsure if this should already be part of the Car prefab.

m4rtint: Place this string on the top, make it read only/get only

PR #16, opened 2019-07-05
Feature/score manager

wbeltran777:
Finished up a simple Score Manager. 
Created another script along with it to check for collisions as if it was a parking spot.


Assets/Script/Managers/ScoreManager.cs
```diff
@@ -0,0 +1,114 @@
...
+/// <summary>
+/// Score Manager keeps track of the score of each Player
+/// Is in charge of incrementing, reseting and initializing the score
+/// </summary>
+public class ScoreManager : MonoBehaviour
+{
+    private static ScoreManager m_instance;
+    private static Dictionary<string, int> m_carScores = new Dictionary<string, int>();
+
+    [SerializeField] private int WinningScore = 2;
```
m4rtint: m_winningScore

Assets/Script/Managers/ScoreManager.cs
```diff
@@ -0,0 +1,114 @@
...
+            // Probably add a call to StateManager here where we can pause the game.
+            Debug.Log(carThatScored.name + " Wins!");
+        }
+    }
+
+    /// <summary>
+    /// Initializes the score of all the cars in Play. Sets them to 0
+    /// </summary>
+    /// <param name="carsInPlay"> List of Cars in Play </param>
+    public void ScoreSetup(List<GameObject> carsInPlay)
```
m4rtint: Might???? be safer, if you require the argument to require List<CarController> instead of GameObject.

So other devs can ONLY pass in cars instead of other random gameobjects, to make sure you're passing in the correct thing.

m4rtint: it's not a big deal

Assets/Script/Managers/WorldManager.cs
```diff
@@ -21,13 +21,18 @@ public class WorldManager : MonoBehaviour
 
     private readonly string namePrefix = "Player";
     private List<Transform> m_carSpawnPoints = new List<Transform>();
+
+    private List<GameObject> m_carsInPlay = new List<GameObject>();
```
m4rtint: I should already have this in WorldManager
You'll need to merge from develop branch

Assets/Script/Managers/ScoreManager.cs
```diff
@@ -0,0 +1,114 @@
...
+            Destroy(gameObject);
+        }
+
+        DontDestroyOnLoad(gameObject);
+    }
+
+    /// <summary>
+    /// Sets the score of every player to 0. Likely used when retrying rounds
+    /// </summary>
+    private void ResetCarScoreForRetry()
```
Gursaanj: Couldn't this function Just call ScoreSetup again?  Just Wondering

m4rtint: Good point.
But I think i would do like
void ScoreSetup(List<GameObject> carsInPlay) 
{
m_currentCarsInPlay = carsInPlay;
ResetCarScoreForRetry();
}

and remove the foreach loop

Gursaanj: Ooh I see what you mean, that sounds good

Assets/Script/Managers/WorldManager.cs
```diff
@@ -16,11 +16,23 @@ public class WorldManager : MonoBehaviour
...
+    private readonly string namePrefix = "Player";
+    private List<Transform> m_carSpawnPoints = new List<Transform>();
+
+    private List<GameObject> m_carsInPlay = new List<GameObject>();
+
+    [SerializeField] private SplitScreenManager m_splitScreenManager;
+    [SerializeField] private ScoreManager m_scoreManager;
+    [SerializeField] private int m_numOfPlayers;
+    [SerializeField] private GameObject m_carTemplate;
+=======
```
m4rtint: You pushed some merge conflicts here. :(

Assets/Script/Managers/ScoreManager.cs
wbeltran777: I'm not sure what you mean by require a list instead of GameObject since the argument is already a list. How could I add the restrictions you're talking about?

Assets/Script/Managers/WorldManager.cs
wbeltran777: Yep my bad, next push should get rid of them

Assets/Script/Managers/ScoreManager.cs
m4rtint: So like
public void ScoreSetup(List<CarController> carsInPlay) {}

so that the only objects you can feed in, are cars (objects with the Car Controller script attached).. Not random GameObjects.

Assets/Script/Managers/WorldManager.cs
```diff
@@ -16,11 +16,20 @@ public class WorldManager : MonoBehaviour
     private readonly string Player = "Player_";
     private readonly string Camera = "Camera_";
 
+
+    private readonly string namePrefix = "Player";
+
+    private List<GameObject> m_carsInPlay = new List<GameObject>();
+
+    [SerializeField] private ScoreManager m_scoreManager;
+    [SerializeField] private int m_numOfPlayers;
```
m4rtint: Remove this, m_numberOfPlayers is already on line 40.

Assets/Script/Managers/WorldManager.cs
```diff
@@ -16,11 +16,20 @@ public class WorldManager : MonoBehaviour
     private readonly string Player = "Player_";
     private readonly string Camera = "Camera_";
 
+
+    private readonly string namePrefix = "Player";
+
+    private List<GameObject> m_carsInPlay = new List<GameObject>();
```
m4rtint: remove this field. 

Already have m_listOfCars; 

If you need a list, you can create a getter to get the Array of m_listOfCars as a List.

Assets/Script/Managers/WorldManager.cs
```diff
@@ -16,11 +16,20 @@ public class WorldManager : MonoBehaviour
     private readonly string Player = "Player_";
     private readonly string Camera = "Camera_";
 
+
+    private readonly string namePrefix = "Player";
```
m4rtint: Remove this since we already have 
readonly string Player = "Player_"

Assets/Script/Managers/WorldManager.cs
m4rtint: hmm..I see you change m_listOfCars to use your 
m_carsInplay. 

You should remove m_listOfCars then, instead of this.

Assets/Script/OnTriggerEnterTest.cs
```diff
@@ -0,0 +1,24 @@
+﻿//-----------------------------------------------------------------------
```
m4rtint: Should delete this file if it's just a test and not needed

Assets/Script/Managers/WorldManager.cs
wbeltran777: I already turned m_carsInPlay into a getter that grabs m_listOfcars, is that alright?

Assets/Script/Managers/ScoreManager.cs
wbeltran777: Hmmm I'm sorry I still don't understand, 
List needs an argument in it's brackets, otherwise I get an error. 
Wouldn't we need to define a car struct or something if we want to only take in cars?

m4rtint: I was just pseudo coding. 

public void ScoreSetup(List<GameObject> carsInPlay){} 

As a developer, I can pass in 10 empty gameobjects into the ScoreSetup and it will think we have 10 cars.
As a developer, I can pass in 5 Obstacles into the ScoreSetup and it will think we have 5 cars.

BUT
 
If i made it
public void ScoreSetup(List<CarController> carsInPlay) {...}

As a developer, I can TRY to pass in 10 empty gameobjects, but Visual Studios/Unity will complain
As a developer, I can TRY to pass in 3 obstacles, but Visual Studios/Unity will complain. 
As a developer, I can ONLY pass in 2 cars (Which contains CarController scripts), and that is the only argument that works. 

It make it so that ScoreSetup doesn't have to worry whether or not what you are passing in are Cars or not.

Assets/Script/Managers/WorldManager.cs
```diff
@@ -72,12 +87,17 @@ private void SetupPlayersAndCameras()
...
     {
         GameObject newPlayer = new GameObject();
         newPlayer.transform.position = m_carSpawnPoints[playerNumber].position;
         newPlayer.name = Player + (playerNumber + 1);
 
+        m_carsInPlay.Add(newPlayer);
+
         GameObject playerCar = CreateCar(m_carSpawnPoints[playerNumber]);
         playerCar.transform.parent = newPlayer.transform;
         m_listOfCars[playerNumber] = playerCar;
```
m4rtint: remove
```
private GameObject[] m_listOfCars
m_listOfCars[playerNumber] = playerCar
```

And just keep m_carsInPlay as its own variable

```
private List<GameObject> m_carsInPlay
public List<GameObject> CarsInPlay { get; }
```

Assets/Script/Managers/ScoreManager.cs
wbeltran777: Yeah that makes sense. However if I make a list of CarController then ScoreSetup loses access to the name, which it uses to setup the dictionary for keeping track of the score, and then to announce the winner. Unless you're suggesting we change the entire indexing of scores to be attached to CarController instead of GameObject. 

We could do that, but every time we park somewhere, we'd have to use GetComponent in order to access a car's CarController. 
Would this be alright? 

How about we keep the GameObject and just throw an error if the object passed doesn't contain a CarController component?

m4rtint: You can access the gameObject by doing 

CarController.gameObject

m4rtint: You don't have to change it to this if you don't want to.

mikeslgoh:
[Score Manager](https://trello.com/c/PSa7aMNd/33-score-manager)

PR #19, opened 2019-07-06
Feature/controller input

mikeslgoh:
Will allow for 2 PC players max but up to 4 controllers attached. Only works for PS4 controllers and there is a bug that has been logged to improve for future iteration


Assets/Script/Managers/WorldManager.cs
```diff
@@ -74,41 +76,58 @@ private void FixedUpdate()
     private void SetupPlayersAndCameras()
     {
         CheckNumberOfPlayers();
+        GameObject[] carSpawnPointsContainer = GameObject.FindGameObjectsWithTag("Spawnpoint");
```
m4rtint: TagName.m_spawnpoint

mikeslgoh:
[Multiplayer - controller input](https://trello.com/c/GXnQyQXN/21-multiplayer-controller-input)

PR #24, opened 2019-07-07
Created score display script. Make sure display component is attached…

wbeltran777:


Assets/Script/DisplayScoreData.cs
```diff
@@ -0,0 +1,60 @@
...
+        //UpdateScoreDisplay();
+    }
+
+    public void UpdateScoreDisplay(string PlayerName)
+    {
+        player1_Score = ScoreManager.CarScores["Player_1"];
+        player2_Score = ScoreManager.CarScores["Player_2"];
+        player3_Score = ScoreManager.CarScores["Player_3"];
+        player4_Score = ScoreManager.CarScores["Player_4"];
+        textComponent.text = "Player 1: " + player1_Score + "\n" +
```
spicyroboto: Looked over this with Martin, and we aren't really sure how this is going to render to the screen in the correct locations as it seems just be one giant string. Is this just showing all the scores in a row somewhere on the screen?

Assets/Script/Managers/ScoreManager.cs
```diff
@@ -17,10 +17,16 @@ public class ScoreManager : MonoBehaviour
     private static ScoreManager m_instance;
     private static Dictionary<string, int> m_carScores = new Dictionary<string, int>();
 
+    public delegate void ScoreDelegate();
+
+    public event ScoreDelegate onScoreChange;
```
spicyroboto: Remove if not being used.

Assets/Script/DisplayScoreData.cs
wbeltran777: Yup, this script will attach to the a scoreDisplay prefab which already has the text aligned at the top right corner, this script just updates that text. 

This was a quick score display for the day of though. I can revise this and make something more tailored to each player screen.

m4rtint: Would this break if there are less than 4 players?

Use String format, it's a lot more cleaner
https://docs.microsoft.com/en-us/dotnet/api/system.string.format?view=netframework-4.8

wbeltran777: It wouldn't necessarily break, but if only two players are playing, you'd still see player 3 and 4 scores as 0.
But yeah it's very rough, we can reject this pull and I can work on it a bit more.

m4rtint: No need to reject it, just change it here.

Assets/Scenes/ParkingLot_Gazeebo.unity
```diff
@@ -6430,6 +6435,72 @@ Transform:
     type: 3}
   m_PrefabInstance: {fileID: 1787937704}
   m_PrefabAsset: {fileID: 0}
+--- !u!1 &1830604527
+GameObject:
```
mikeslgoh: Are you working of the Gazeebo scene? Might want to create a new scene

Assets/Script/CarController.cs
```diff
@@ -81,7 +81,7 @@ private void CarControllerOnUpdate()
         float vertical;
         if (!m_playerController.Equals(ControllerPlatform.PC))
         {
-            Debug.Log(m_playerID);
+//            Debug.Log(m_playerID);
```
mikeslgoh: Once fully tested, you can remove your Debug.Logs

Assets/Script/DisplayScoreData.cs
mikeslgoh: I agree with Martin to use string format since it would just make it look cleaner. If you have a way of getting the number of players, you can just concat the strings into one variable and do something like:

> 
```
scoreText = "";
for (int i=1; i <= numberOfPlayers; i++){
     playerID = "Player_"+i;
     playerScore = ScoreManager.CarScores[playerID];
     scoreText.concat( --- enter your string here --- );
}
// set scoreText for UI Text Component
```

spicyroboto: Upvoting Micheal’s comment.

wbeltran777: I'm working on it right now, and I'm thinking to modify SplitScreenManager to prefabed Canvas and Text to each camera. Like

prefab.parent=playerCamera

And so I could use a SerializeField in SplitScreenManager. 

Would this be too much tinkering with existing scripts or is that alright? I can also try to keep this more modular.

m4rtint: Test it, see if it works first. 

Then refactor. Keep us updated 👍

ProjectSettings/DynamicsManager.asset
```diff
@@ -28,3 +28,6 @@ PhysicsManager:
     m_Center: {x: 0, y: 0, z: 0}
     m_Extent: {x: 250, y: 250, z: 250}
   m_WorldSubdivisions: 8
+  m_FrictionType: 0
```
m4rtint: How come there are changes in this file?

Assets/Script/PointsUpdater.cs
```diff
@@ -0,0 +1,26 @@
+﻿using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+using UnityEngine.UI;
+
+public class PointsUpdater : MonoBehaviour
```
m4rtint: This isn't used anywhere

Assets/Script/Managers/ScoreManager.cs
```diff
@@ -53,13 +53,25 @@ public void CarScored(GameObject carThatScored)
     {
         // We can modify this later with time values
         m_carScores[carThatScored.name]++;
+        carThatScored.GetComponentInChildren<DisplayPointsHandler>().UpdateCarScoreDisplay(carThatScored.name);
```
m4rtint: You shouldn't need to do this. 
You should set a delegate in ScoreManager. 
Whenever someone scores, ScoreManager would call a delegate

Assets/Script/DisplayPointsHandler.cs
```diff
@@ -0,0 +1,29 @@
...
+    private string m_scoreDisplayString;
+    
+    /// <summary>
+    /// Updates score every time a Car Scores
+    /// </summary>
+    /// <param name="carName"> Name of current car that just scored </param>
+    public void UpdateCarScoreDisplay(string carName)
+    {
+        m_scoreDisplayString = string.Format("Player {0} Score: {1}",carName.Substring(carName.Length-1,1),m_currentCarScore);
+        m_currentCarScore++;
```
m4rtint: why do you need a score here as well? can't you just grab it from the score manager?

Assets/Script/Managers/ScoreManager.cs
wbeltran777: I'm trying that out, but if I subscribe UpdateCarScoreDisplay to a Score Manager event, then all of the car's UpdateCarScoreDisplay will get called. So all players end up with the same points. So when a car scores I need to make sure that only the car that scored is subscribed, which would require me to access their DisplayPointsHandler class anyway. Alternatively I could make four separate events, one for each player and call each event depending on which player scores. What do you think?

m4rtint: yeah you're right, need to figure that out.

m4rtint: I think it's fine for now, but since we're going to reorganize and refactor some stuff, this solution is alright for now.

ProjectSettings/DynamicsManager.asset
wbeltran777: I may have messed around with some controller stuff and forgot about it, how can I revert these particular changes? Same goes for some of the scene modifications.

m4rtint: git checkout origin/develop {filename}

m4rtint:
[Display score UI](https://trello.com/c/JosMYXoz/48-display-score-ui)

m4rtint:
[Display score UI](https://trello.com/c/JosMYXoz/48-display-score-ui)

Assets/Script/DisplayPointsHandler.cs
wbeltran777: I'll pass it in as a parameter instead

Assets/Script/PointsUpdater.cs
wbeltran777: it's dead now

ProjectSettings/DynamicsManager.asset
wbeltran777: Removed changes from scenes and project settings I changed

PR #27, opened 2019-07-11
Controller selection setup

mikeslgoh:
Created a test game flow where the user would select how many players (and what controls they would be using) and then hit START and be taken to the game scene.

**NOTE**: Menu is definitely not set in stone in terms of design/layout and I needed to do this in order to fix a bug I was encountering in my previous implementation. Would like some feedback before eventually merging this and will StyleCop :)


mikeslgoh:
[Controller Selection Scene](https://trello.com/c/ZFYqyTsg/39-controller-selection-scene)

Assets/Script/ControllerSelection.cs
```diff
@@ -0,0 +1,151 @@
...
+            {
+                numberOfPS++;
+                PlayerInformation playerInfo = new PlayerInformation(numberOfPS, i+1, ControllerPlatform.PS);
+                playerInfoList[i] = playerInfo;
+            }
+        }
+        GameData.PlayerInformationList = playerInfoList;
+    }
+
+    private bool CheckNumberOfControllers()
```
m4rtint: are you checking the number of PC/PS inside two different methods?

Assets/Script/Managers/StateManager.cs
```diff
@@ -102,10 +102,13 @@ public bool IsPaused()
     public void PauseGame()
     {
         m_state = State.Paused;
-        PauseOnSceneRigidBody();
-        if (OnPause != null)
+        if (m_listOfRigidBodies != null)
```
m4rtint: How come you added this?

Assets/Script/Managers/WorldManager.cs
```diff
@@ -72,6 +72,10 @@ private void Start()
         SetupPlayersAndCameras();
         SetNewParkingSpot();
         m_scoreManager.ScoreSetup(m_carsInPlay);
+
+        m_numberOfPlayers = GameData.PlayerCount;
+        
+        StateManager.Instance.PlayGame();
```
m4rtint: Can you write //DEBUG, since this should be temporary for now

Assets/Script/ControllerSelection.cs
mikeslgoh: Yeah I am. One was to check how many and the assertions. The other is to assign which player is a PC or PS. I might work to combine them together and just have the assertion checks in the above function.

Assets/Script/Managers/StateManager.cs
mikeslgoh: this was throwing a null pointer error until I added the check

Assets/Script/Managers/WorldManager.cs
mikeslgoh: On line 77?

m4rtint: yeah

Assets/Script/Managers/StateManager.cs
m4rtint: Hmm strange, didn't think it'll do that.

Assets/Script/ControllerSelection.cs
m4rtint: okay

Assets/Script/CarController.cs
```diff
@@ -89,12 +85,12 @@ private void CarControllerOnUpdate()
         float vertical;
         if (!m_playerController.Equals(ControllerPlatform.PC))
```
m4rtint: u can probably pass m_playerController into InputManager, and let it handle everything there?

mikeslgoh: yeah i'll try and play around with this too.

PR #30, opened 2019-07-16
Change default GameData to have one player

mikeslgoh:
The single scenes were breaking because there was no default GameData for player information. I have initialized 1 player to be created for each scene so that it does not rely on the menu flow.


mikeslgoh:
[(2) Original maps broken [0]](https://trello.com/c/jLrTRIo0/58-2-original-maps-broken-0)

PR #31, opened 2019-07-17
Feature/level selector

Gursaanj:
Created a Level Selector in the Controller Setup Scene (Temporary Menu Scene) where the player can select the map they want to play in. Also added a random level selector to go with it. 

image


Assets/Script/ControllerSelection.cs
```diff
@@ -28,6 +34,27 @@ void Start()
...
+    // have another script with all scene names attached
+    private void SetLevels()
+    {
+        // Current Issue is that Build Settings automatically puts the controller Selection scene on the end of the list of scenes
+        // so its being counted with with SceneCount Function. Temporary fix: -1 
+        m_numberOfLevels = SceneManager.sceneCountInBuildSettings -1;
+        m_listOfScenes = new List<string>();
+
+        for (int i = 0; i < m_numberOfLevels; i++)
+            m_listOfScenes.Add(System.IO.Path.GetFileNameWithoutExtension(SceneUtility.GetScenePathByBuildIndex(i)));
```
m4rtint: anything with IO and system it slow.  Usually you would stay far away from Files unless you must must must need it

Get the scene name?
https://docs.unity3d.com/ScriptReference/SceneManagement.Scene-name.html

Assets/Script/ControllerSelection.cs
```diff
@@ -28,6 +34,27 @@ void Start()
...
+
+        LevelControlSelection.ClearOptions();
+        LevelControlSelection.AddOptions(m_listOfScenes);
+        LevelControlSelection.AddOptions(new List<string> {"Random Level"});
```
m4rtint: Put the string as its own variable.

Assets/Script/ControllerSelection.cs
```diff
@@ -10,16 +10,22 @@ public class ControllerSelection : MonoBehaviour
     [SerializeField] private Text numberOfPlayers;
     [SerializeField] private Text errorMessage;
     [SerializeField] private Dropdown playerControlSelection;
+    [SerializeField] private Dropdown LevelControlSelection;
```
m4rtint: m_levelControlSelection

m4rtint: @mikeslgoh leading by example i see 🥇

Assets/Script/ControllerSelection.cs
Gursaanj: Or should I  make List<String> variable  m_OtherLevels, so we can add all special levels (if we have them) into one list ?

Assets/Script/ControllerSelection.cs
Gursaanj: Should I just change all the variable into m_" " ?

Assets/Script/ControllerSelection.cs
m4rtint: Nah it's fine for now. 
just make a readonly string for this "Random Level"

Assets/Script/ControllerSelection.cs
m4rtint: if u want to do code clean up, sure

Assets/Script/ControllerSelection.cs
```diff
@@ -56,16 +85,24 @@ public void Decrement()
...
     }
 
-    // If the number of controllers are correct, save the players' information and start the game
+    // If the number of controllers are correct, save the players' information and start the game with given Level
     public void StartGame()
     {
         if (SavePlayerData())
         {
-            SceneManager.LoadScene("Scenes/LevelFlow_Testing/Game");
+            if (LevelControlSelection.options[LevelControlSelection.value].text == "Random Level")
```
m4rtint: this still uses the string

Gursaanj: yup Im just an idiot haha one sec

m4rtint: A big thing that developers always tell you to do. 
Never do string comparison - for dev safety reasons. 

you would normally use Enums

Gursaanj:
> Scene changing stuff should not be in the controller selection.
> 
> Scene changing stuff should have its own script that handles scene changing stuff.

@m4rtint 
If that is the case, Should I put the start game function in its own script as well??, Since the Controller Selection stuff??

Assets/Script/ControllerSelection.cs
Gursaanj: https://issuetracker.unity3d.com/issues/scenemanager-dot-getscenebybuildindex-dot-name-returns-an-empty-string-if-scene-is-not-loaded

Gursaanj: I cant get the names from GetSceneFromBuildIndex().name if the scene isnt loaded (I just get nulls)
So current solution would be too keep it as it is for now.
Or create a function takes the path of the Scene (from the SceneUtility thingy) and get substring of the path that is the name, and then use that 

Which do you think is more efficient?

wbeltran777: How about
` List<string> scenes = new List<string>();
 foreach(EditorBuildingSettingsScene scene in EditorBuildSettings.scenes)
 {
     if(scene.enabled)
         scenes.add(scene.path);
 }`

or looks like SceneManager can let you get the scenes
`int countLoaded = SceneManager.sceneCount;
        Scene[] loadedScenes = new Scene[countLoaded];
 
        for (int i = 0; i < countLoaded; i++)
        {
            loadedScenes[i] = SceneManager.GetSceneAt(i);
        }`

Gursaanj: Hmm since sceneCount and GetSceneAt only access loaded scenes ,I dont think it retrieves the scenes we need, which aren't loaded

m4rtint: In my opinion, u shouldn't find the list of scene from the scene editor. 
You should have a  container/class/data structure/entity that contains 
1) Name of the scene that you want to display to the user
2) name of the scene file name. 
3) Some enum somewhere that links this data structure together

So that you do like

// which returns the "The wtvr gazibo"
SceneManager.GetSceneName(Enum.Gazibo)

//Which returns string of scene file name.
SceneManager.GetSceneFileName(Enum.WaterFountain)

something a lot more contained.

Gursaanj: Okay Will try to do that some time tonight. Have never used Enum before or created a struct so will need to familiar myself first. But for now (I.e. like 1 min you made the comment haha) I created a function that gets the scene name from the scene path, so at least there is something temporary

Assets/Script/ControllerSelection.cs
```diff
@@ -12,11 +12,15 @@ public class ControllerSelection : MonoBehaviour
     [SerializeField] private Dropdown playerControlSelection;
     [SerializeField] private Canvas canvas;
     [SerializeField] private float dropdownShift = -20;
+    [SerializeField] private Dropdown m_LevelSelector;
```
m4rtint: Reminder to remove all this stuff from the controller selection

Gursaanj: Hmm the only reason that is still in there is for the start game function. Just like the comment above, Should I just put the start game function in its own script (Which makes the most sense to me). 

So then the controller setup stuff only takes the controller setup and the level selection stuff only takes the level selection stuff, and both pass info into the start game script

Assets/Script/ControllerSelection.cs
m4rtint: in a nutshell
Enum - List of things you're limiting the user/devs to pick from. 
E.g. Colors, Directions, Weapons...etc

Struct - A simple object that stores what you want.

Enum Color {
Blue,
Green,
Red
}

struct Box {
Enum color;
int width;
int height;
}

So now I can do 
Box someBox = new Box (Color.Green, 10, 100)
someBox.color;
someBox.width;
someBox.height;

Assets/Script/ControllerSelection.cs
m4rtint: yes

Assets/Script/LevelSelection.cs
```diff
@@ -0,0 +1,73 @@
...
+        m_levelSelector = GetComponent<Dropdown>();
+        SetLevels();
+    }
+
+
+    //Takes the level names from the Build Settings and makes them option in the Level Selector Dropdown
+    // *** Will only work if all the levels in the Build Settings are actual Playable Scenes (I.e. No Menu/Settings Scenes) ***
+    // A way to combat this issue would be to adopt a naming convention so we can weed out the scenes we want / not want or
+    // have another script with all scene names attached
+    private void SetLevels()
```
m4rtint: call this level initializer or something. 

Cuz it could be confused as a setter

Assets/Script/ControllerSelection.cs
m4rtint: Reminder to remove this stuff

Assets/Script/ControllerSelection.cs
```diff
@@ -97,7 +111,7 @@ private bool SavePlayerData()
                     PlayerInformation playerInfo = new PlayerInformation(numberOfPC, i+1, ControllerPlatform.PC);
                     playerInfoList.Add(playerInfo);
                 } catch (AssertionException e) {
-                    errorMessage.text = "Too many PC players added";
+                    m_errorMessage.text = "Too many PC players added";
```
m4rtint: We'll soon need to place this stuff into a separate file @mikeslgoh

Assets/Script/ControllerSelection.cs
Gursaanj: Haha yeah each commit I do is to tackle one of the comments at a time. Just got to it. Now for the Enum stuff

Assets/Script/ControllerSelection.cs
mikeslgoh: The issue with using Editor functions is that there is no guarantee that it will work when we finally build it

Assets/Script/ControllerSelection.cs
mikeslgoh: I did say I wanted to StyleCop everything again :/

Assets/Script/ControllerSelection.cs
mikeslgoh: Yeah at work I use localized strings so is that what you are thinking of doing?

m4rtint: something similar to the TagName script

Assets/Script/LevelSelection.cs
```diff
@@ -0,0 +1,78 @@
...
+using System.Collections.Generic;
+using UnityEngine.SceneManagement;
+using UnityEngine.UI;
+using UnityEngine;
+
+public class LevelSelection : MonoBehaviour
+{
+    private Dropdown m_levelSelector;
+    private int m_numberOfLevels;
+    private List<string> m_listOfScenes;
```
m4rtint: as far as i can tell, this is only needed inside Levelintializer() 

Just create it locally inside the method. No need to initialize it here.

Assets/Script/LevelSelection.cs
```diff
@@ -0,0 +1,78 @@
...
+        // so its being counted with with SceneCount Function. Temporary fix: -1 
+        m_numberOfLevels = SceneManager.sceneCountInBuildSettings - 1;
+        m_listOfScenes = new List<string>();
+
+        for (int i = 0; i < m_numberOfLevels; i++)
+        {
+            //m_listOfScenes.Add(System.IO.Path.GetFileNameWithoutExtension(SceneUtility.GetScenePathByBuildIndex(i)));
+            m_listOfScenes.Add(GetSceneName(SceneUtility.GetScenePathByBuildIndex(i)));
+        }
+
```
m4rtint: you can do 
```m_listOfScenes.Add(m_randomLevel);```
here 
instead of doing AddOptions Twice.

Looks cleaner

**Comment by Gursaanj** [comment]
[Add random level selector before beginning the game](https://trello.com/c/MxEyvpCp/56-add-random-level-selector-before-beginning-the-game)

**Comment by Gursaanj** [comment]
[Add random level selector before beginning the game](https://trello.com/c/MxEyvpCp/56-add-random-level-selector-before-beginning-the-game)

File: Assets/Script/UI/StartGame.cs [hunk 10]
```diff
@@ -0,0 +1,38 @@
...
+
+    /// <summary>
+    /// The Level Selector to know which scene the user wants to load
+    /// </summary>
+    [SerializeField] private Dropdown m_levelSelector;
+
+    /// <summary>
+    /// When the Player data is sorted, load the scene selected in the LevelSelector 
+    /// </summary>
+    public void StartTheGame()
```
wbeltran777: What's the benefit of having a start game class instead of just starting the game from the level selector class?

Gursaanj: Its really more of a flow thing . Since right now we only need the Controller input, and the level selection information to start the game, so I guess its fine to not have a start class. But in the future, if more setup info is required, then it looks cleaner (in my opinion) for all information to be passed to one start script.

PR #33, opened 2019-07-25
Feature/a star ai

Gursaanj:
NPC and an NPC Manager has been set up and now any number of NPC in a scene will be given a list of parking spots (all parking spots split among the NPC) that they will use of list of destinations to go to. 

The NPC is far from polished, (not clean, still gets stuck at the most random things, and no avoidance as of yet) but at least it gives us a small sense of what they will be doing, when we test out the game from now onwards. 

The 3 scenes (Gazeebo, Office and Fountain) have been duplicated, in which they have the AI added in them. These scenes have been updated on the Level Selector as well. 
Gotty Animation


Gursaanj:
[Multiple NPC working cohesively in one scene](https://trello.com/c/vsq5m6x4/38-multiple-npc-working-cohesively-in-one-scene)

PR #36, opened 2019-07-28
Modified car physics, added break button and hand break

wbeltran777:
Break button works on PC, it should just be left control for testing purposes. I think we discussed that we'd leave all the specific platform specific input stuff for Michael to do (sorry Michael).

The Parking Button is also implemented, but commented out. I was trying it out but I think it'd be best to try it out with the whole team and see how we feel. 

Similar thoughts go to the car physics. I did decrease the bounciness a bit so if you run into a wall you won't just immediately ricochet away. Also modified the m_maxStickyVelocity which basically makes the car have more right/left ward movement if the car is turning too quickly.


m4rtint:
CarControlPolish.unity 

is this scene needed?

Assets/Script/Game/Car/CarController.cs
```diff
@@ -106,6 +112,13 @@ private void CarControllerOnUpdate()
 
         // Keeps motion only in the forward direction, with some right direction to account for drifting. 
         m_rigidbody2D.velocity = ForwardVelocity() + (RightVelocity() * m_currentDriftFactor);
+
+        // If loop for Hand Break Mechanic
```
m4rtint: Remove if not needed

Assets/Script/Game/ParkingDetection.cs
```diff
@@ -108,6 +108,8 @@ private bool IsParked(Collider2D collision)
         bool isTagCorrect = collision.gameObject.tag == TagNames.CarTag;
         bool isParked = IsParked(collision, GetComponent<Collider2D>());
 
-        return isParked && isTagCorrect && m_isChosenToBeParkingSpot;
+        return 
+            // InputManager.HandBreak() &&
```
m4rtint: remove if not needed

mikeslgoh:
[Add break button for keyboard](https://trello.com/c/HWnnQzbK/67-add-break-button-for-keyboard)

PR #37, opened 2019-07-28
Feature/parking indicator lerping

wbeltran777:
I'm making the colors go back and forth between orange and blue to kinda of work with the orange and blue thing we got going on with the menus possibly. It also makes the parking spot pop out more. Let me know what yall think!


Assets/Script/Game/ParkingDetection.cs
```diff
@@ -19,6 +19,9 @@ public class ParkingDetection : MonoBehaviour
     [SerializeField]
     private GameObject m_glowingSpot;
 
+    private readonly Color m_orange = new Color(1,224/255,173/255);
+    private readonly Color m_blue = new Color(192 / 255, 173 / 255, 1);
```
m4rtint: Stylecop

PR #38, opened 2019-07-29
Feature/pause game

Gursaanj:
Created a pause panel (like the end game panel) where if a player were to press the pause button the pause menu would appear. For this case, this has bee implemented for a single PC player only. This has not bee tested for controllers or multiple players. 

This is how it works 

Gotty working Pause

However, I have reached a couple of issues and possibly a bug (or something I dont understand). 

First issue is that for now, if I assign the same button to be the Pause button (when the game is in play) and the Resume button (when the game is paused) then it only works for 50% of the time. Not sure if it is how I am calling the buttons but resuming the game using the same button as pausing it is really clunky and basically unusable - In the InputManager

The bigger problem is that if I start from the Level Selection scene (where we decide the controllers and the level) and choose a level, then in the game reset the level (as shown below) I get that null exception . Also I notice that the StateManager is gone.  So when I pause the game then, the Pause panel does not come up and the rigidbody freezes

Gotty broken Pause

The state manager also goes missing if we go from the end game panel to the level selector to another scene.


Gursaanj:
[Update UI - PAUSE game](https://trello.com/c/oxP9p8zI/73-update-ui-pause-game)

Assets/Scenes/LevelFlow_Testing/Controller_Selection.unity
```diff
@@ -461,6 +461,11 @@ PrefabInstance:
       propertyPath: m_LocalEulerAnglesHint.z
```
m4rtint: What changes did u do with this scene?

Assets/Script/Managers/InputManager.cs
```diff
@@ -83,6 +83,24 @@ private static float VerticalController(ControllerPlatform controllerPlatform, i
         return 0;
     }
 
+    public static void PauseTheGame()
```
m4rtint: Not an issue, but there will be a time where we have to fix the bug where a player spams the pause key.

Assets/Script/Managers/StateManager.cs
```diff
@@ -102,7 +102,7 @@ public bool IsPaused()
     public void PauseGame()
     {
         m_state = State.Paused;
-        if (m_listOfRigidBodies != null)
+        if (ListOfRigidBodies != null)
```
m4rtint: why is this changed?

Assets/Script/UI/EndGamePanel.cs
```diff
@@ -40,13 +40,15 @@ private void OnDestroy()
...
-    {
-        SceneManager.LoadScene(SceneManager.GetActiveScene().name);
-    }
+    // Will delete the following if everyone is okay with using base panel methods
 
-    private void OnMainMenuButtonpressed()
-    {
-        SceneManager.LoadScene(SceneInformation.MainMenuFileName);
-    }
+    //private void OnReplayButtonPressed()
```
m4rtint: Remove if not needed

Assets/Script/UI/BasePanel.cs
```diff
@@ -37,4 +38,24 @@ protected virtual void Minimize()
...
+    {
+        SceneManager.LoadScene(SceneInformation.MainMenuFileName);
+    }
+
+
+    /// <summary>
+    /// Reload active scene when button is pressed
+    /// </summary>
+    /// <param name="m_scene"></param>
+    protected virtual void OnReplayButtonPressed(string m_scene)
```
m4rtint: Remove the parameter if it's going to be the same argument every single time

Assets/Script/Managers/StateManager.cs
Gursaanj: because m_listofRigidBodies is null until ListOfRigidBodies get function is called (because of the If function)

Assets/Script/Managers/InputManager.cs
Gursaanj: Should have fixed that with new commit

Assets/Script/Managers/InputManager.cs
```diff
@@ -83,6 +83,37 @@ private static float VerticalController(ControllerPlatform controllerPlatform, i
         return 0;
     }
 
+    /// <summary>
+    /// Checks it there is any tweening happening in the scene at the frame called
+    /// </summary>
+    /// <returns></returns>
+    private static bool AnyTweening()
+    {
+        return LeanTween.tweensRunning > 0 ? true : false;
```
m4rtint: I didn't mean to say fix it now. but oh well

just do 
`return LeanTween.tweensRunning > 0`

Assets/Script/Managers/InputManager.cs
```diff
@@ -83,6 +83,37 @@ private static float VerticalController(ControllerPlatform controllerPlatform, i
...
+    }
+
+    /// <summary>
+    /// If the esc key is being called and the panels arent in the middle of popping up or down
+    /// (i.e. no tweening at the moment) then pause the game (if the state is IsPlaying) and 
+    /// resume the game if the state isPaused
+    /// </summary>
+    public static void PauseAndUnpause()
```
m4rtint: This is not good. pause and unpause. others developers would have no idea if it's in an unpause or paused state. 
should just put AnyTween() in the original pause and resume methods

Assets/Script/Managers/WorldManager.cs
```diff
@@ -97,6 +97,10 @@ private void FixedUpdate()
         {
             OnFixedUpdate();
         }
+
+        //Pause the game if the Pause button is pressed Need to place it here so that the resume button works when state is IsPaused
+        InputManager.PauseAndUnpause();
```
m4rtint: I'm thinking maybe put this in Update() {} function

Assets/Script/Managers/InputManager.cs
Gursaanj: Yeah thats what Im thinking (i put that in my commit notes). So there is a bool statment we can put on the pausepanel : LeanTween.istweening(gameobject) which will let us know if the pause panel is tweening (which would be much better) But I am having trouble thinking about how to pass that info from pausepanel to input manager- Any Ideas?

Assets/Script/Managers/WorldManager.cs
Gursaanj: hmmm, I can do that, is that because in this case we dont care about physical based timing or something like that

Gursaanj: Actually you are right, the transition is much smoother, I will place the method in a Update Function

Assets/Script/Managers/InputManager.cs
Gursaanj: Okay the latest commit (Better Tween check for Input Manager) asks only if the pause game panel is tweening (better solution) but in order to do so I had to make some static variables as well as add an onfixedupdate method (to check whether the gameobject is tweening or not).

Gursaanj: The issue with that is that the update for pausegame panel is on a fixed update and the update for the input manager (to pause and unpause the game) is on regular update. Causing the input and bool in question to be updated out of sync. Should we think of having a global update method too??

m4rtint: I just tested it without doing tween checking (spamming escape). it works perfectly fine. 
no bugs. 
could remove this tweening check stuff.

m4rtint: Rename is PauseAndPlay()

PR #40, opened 2019-08-11
Refactor/time scale pause

m4rtint:


m4rtint:
[Refactor Pause to use timescale](https://trello.com/c/zWI90QaR/87-refactor-pause-to-use-timescale)

Assets/Script/Managers/StateManager.cs
```diff
@@ -134,6 +106,7 @@ public void PlayGame()
     public void EndGame()
     {
         m_state = State.EndGame;
+        Time.timeScale = 1.0f;
```
Gursaanj: Should we stop time when the game has ended (i.e. timescale = 0.0f)?

m4rtint: Mmmmmm I thought it'll be funnier if we let it bounce around when the game ends when peolle still accelerating

Gursaanj: Oh I see haha fair enough. But if that is the case, wouldnt the timescale already be 1.0 when Endgame is called.  so there is no point of setting the timescale in the method?

m4rtint: It's a precaution

Gursaanj: fair enough

Assets/Script/Managers/StateManager.cs
```diff
@@ -156,34 +129,4 @@ private void Awake()
 
         DontDestroyOnLoad(gameObject);
     }
-
```
wbeltran777: Where are we freezing and unfreezing the rigidbodies now that they're not in statemanager?

m4rtint: No more freezing or refreezing. No need to.

m4rtint: We're just freezing time.

wbeltran777: Groovy

PR #41, opened 2019-08-15
Feature/sound effects manager

Gursaanj:
Created A Sound Effect Manager (Singleton) that can be used by other things in the scene to play a specific sound. 

Can only be tested in the Scene ParkingLot_GazeeboWithSoundEffects

As of right now, when the player collides with another player, when the player collides with an obstacle, when the pause menu is brought up or down and when a player has parked into a target spot  all trigger a specific sound effect (all temp sound effects taken from the internet). The relative velocity of collision dictate the relative volume of the sound effect

The list of sound effects and the crispness of the sound is still lacking, but for now, it will do for testing purposes


Gursaanj:
[SoundEffectManager](https://trello.com/c/o8bsC06M/79-soundeffectmanager)

m4rtint:
Has this been tested with 2+ cars bumping around?

Assets/Script/Data/SoundEffectData.cs
```diff
@@ -0,0 +1,33 @@
...
+
+    /// <summary>
+    /// A sound effect for when a player hits an obstacle (wall or other)
+    /// </summary>
+    ObstacleCollision,
+
+    /// <summary>
+    /// A sound effect to indicate the press of a UI Button (pause/resume/start game ...)
+    /// </summary>
+    UI,
```
m4rtint: UI is way too generic, rename is to something more specific

Assets/Script/Game/Car/CarController.cs
```diff
@@ -120,4 +125,26 @@ private Vector2 RightVelocity()
...
+    /// Details what happens to the player cars when they collide with objects that are not parking spots
+    /// </summary>
+    /// <param name="collision"></param>
+    private void OnCollisionEnter2D(Collision2D collision)
+    {
+        float m_scaledVolume = ((collision.relativeVelocity.magnitude - m_minRelativeVelocity) / (m_maxRelativeVelocity - m_minRelativeVelocity));
+
+        //For Unlikely Cases
+        m_scaledVolume = (m_scaledVolume < 0) ? 0 : m_scaledVolume;
+        m_scaledVolume = (m_scaledVolume > 1) ? 1 : m_scaledVolume;
```
m4rtint: use Mathf.clamp instead of all these if

Assets/Script/Game/Car/CarController.cs
```diff
@@ -24,6 +24,11 @@ public class CarController : MonoBehaviour
 
     [SerializeField] private float m_maxStickyVelocity = 2.8f;
 
+    //Based on testing with curent game setup
+    private readonly float m_maxRelativeVelocity = 6.1f;
```
m4rtint: relative velocity to what? I'm confused with the naming.

Assets/Script/Game/Car/CarController.cs
```diff
@@ -120,4 +125,26 @@ private Vector2 RightVelocity()
...
+
+        //For Unlikely Cases
+        m_scaledVolume = (m_scaledVolume < 0) ? 0 : m_scaledVolume;
+        m_scaledVolume = (m_scaledVolume > 1) ? 1 : m_scaledVolume;
+
+        if (LayerMask.LayerToName(collision.gameObject.layer) == "Obstacle")
+        {
+            SoundEffectManager.Instance.PlaySoundEffect(SOUNDEFFECT.ObstacleCollision, m_scaledVolume);
+        }
+        if (collision.gameObject.CompareTag("PlayerCar"))
```
m4rtint: There's a tagname class you can use. TagName.PlayerCar. or something like that.

Assets/Script/Managers/SoundEffectManager.cs
```diff
@@ -0,0 +1,78 @@
...
+using UnityEngine;
+
+public class SoundEffectManager : MonoBehaviour
+{
+    private static SoundEffectManager m_instance;
+
+    private static AudioSource m_audiosource;
+
+    [SerializeField]
+    private List<AudioClip> m_listOfSounds;
```
m4rtint: This manager is very similar to how I set it up. 
It'll be a mess if you lose, rearrange, mess with the ordering of this list of sounds on the inspector. 

What I did was create a new private AudioClip for EVERY new sound. and organize it with Headers.

Assets/Script/Managers/SoundEffectManager.cs
```diff
@@ -0,0 +1,78 @@
...
+    /// </summary>
+    /// <param name="m_soundEffect"></param>
+    /// <param name="m_relativeVolume"></param>
+    public void PlaySoundEffect(SOUNDEFFECT m_soundEffect, float m_relativeVolume = 1)
+    {
+        m_audiosource.volume = m_relativeVolume;
+
+        switch (m_soundEffect)
+        {
+            case SOUNDEFFECT.PlayerCollision:
```
m4rtint: You should create a new method for every new sound. 

public void CarCollide()
public void CarColliderObstacle()

This way, if something changes with the enum, you don't have to go around all 10 spots that calls SOUNDEFFECT.UI, and change all the names.

m4rtint: And you can add extra logic inside this sound manager if needed...e.g. the relative velocity, which i'm wondering whether or not should be handled in here.

Assets/Script/Managers/SoundEffectManager.cs
```diff
@@ -0,0 +1,78 @@
...
+        {
+            Destroy(gameObject);
+        }
+
+        DontDestroyOnLoad(gameObject);
+    }
+
+    private void Start()
+    {
+        m_audiosource = GetComponent<AudioSource>();
```
m4rtint: we'll need more real soon for multiple cars. Only question is how many?

Gursaanj: Yeah thats why I left it at one for know, but we definitely need more. We need to know how many sound effects can be triggered at once during a scene and work from there

Assets/Script/Managers/SoundEffectManager.cs
Gursaanj: Yeah was thinking of doing that but was wondering how feasible it would be when there was a huge list of sound effects. But for sure, ill do that in my next commit

Assets/Script/Game/Car/CarController.cs
Gursaanj: so relative velocity takes into account the velocity of both sides in the collision and takes the magnitude of the vector3 (or in this case vector2) that is the summed  up velocity (relative to the center of the collision). So after playing around in all the scenes, at least for objects, this is around the highest relativespeed.magnitdue that can take place

Assets/Script/Managers/SoundEffectManager.cs
m4rtint: When there's a huge list of sound effects, organization is needed more than ever.

Assets/Script/Managers/SoundEffectManager.cs
Gursaanj: Hmm that seems like alot of method that do all the same thing though. Wouldnt be better to be more specific with the SOUNDEFFECT Enum and ensure all cases are switched through?

m4rtint: Both ways works. 
I just don't want to down the line, change that one enum name in 25 spots. 

Also, you never know what extra logic you might need when calling for sound. E.g. fade in/out. I'm not the end all be all final decision. So stick with whatevr you think is best.

Gursaanj: Yeah good point, I think for now this should work but when we decide the number of parameters we need and which Enums needs what parameters, then we can make the corresponding functions!!

PR #42, opened 2019-08-17
Feature/major refactoring

mikeslgoh:
- Created BaseWorldManager
- Moved player related functions to PlayerManager
- Renamed scene names as well as updated for SceneData

NOTE: Have not stylecopped yet so please forgive me :')

PR #44, opened 2019-08-19
Feature/scene transition

m4rtint:
Branched from: https://github.com/gotty2019/GOTTY/pull/42


Gursaanj:
[Scene Transition](https://trello.com/c/jXWCF7PH/101-scene-transition)

Assets/Script/UI/EndGamePanel.cs
```diff
@@ -39,4 +39,14 @@ private void OnDestroy()
     {
         StateManager.OnEndGame -= Expand;
     }
-}
\ No newline at end of file
+
+    private void OnReplayButtonPressed()
```
Gursaanj: Why do you recreate the OnReplay and OnMainMenu methods again and not derive them from the BasePanel?

m4rtint: ur correct, will change

m4rtint: Changed

Gursaanj:
Looks good to me. Is there still a reason that this is under the "Do Not Merge" Label?

mikeslgoh:
> Looks good to me. Is there still a reason that this is under the "Do Not Merge" Label?

My code needs to be merged before this which is why this has the label still

m4rtint:
Also Fixed the AI Scenes

PR #47, opened 2019-08-23
Bugfix/incontrolsfx

m4rtint:


m4rtint:
Fixes:
InControl only has one singleton at a time
border doesn't have car overlapping
StyleCopped
SFX Manager added into scenes

m4rtint:
[Car looks weird between borders](https://trello.com/c/3HLZ8QKH/94-car-looks-weird-between-borders)

Assets/Script/UI/PlayerUserInterface.cs
```diff
@@ -38,6 +38,7 @@ public void Initialize(Camera camera, Color color)
     {
         GetComponent<Canvas>().worldCamera = camera;
         GetComponent<Canvas>().planeDistance = 1;
+        GetComponent<Canvas>().sortingLayerName = TagNames.UILayer;
```
m4rtint: This is so that the border overlaps everything else on the screen (including car)

Assets/Script/Managers/SoundEffectManager.cs
```diff
@@ -83,4 +91,23 @@ public void PlaySoundEffect(SOUNDEFFECT m_soundEffect, float m_relativeVolume =
                 break;
         }
     }
+
+    private void Awake()
```
m4rtint: Singleton SFXManager so it appears throughout the scenes

Assets/InControl/Source/Components/InControlManager.cs
```diff
@@ -49,6 +49,18 @@ void OnEnable()
 			}
 		}
 
+        private void Awake()
```
m4rtint: Making InControl manager singleton

Assets/InControl/Source/InputManager.cs
```diff
@@ -127,7 +127,7 @@ static void AssertIsSetup()
 		{
 			if (!isSetup)
 			{
-				throw new Exception( "InputManager is not initialized. Call InputManager.Setup() first." );
+                Setup();
```
m4rtint: Instead of throwing exception of setup problems, i just force it to setup if it's not setup

Assets/Script/UI/PlayerUserInterface.cs
Gursaanj: would just layering the border in front of the parking lots work as well?

Assets/Script/Managers/SoundEffectManager.cs
Gursaanj: Do you want to put the audio clips via Resources.load stuff here ?

m4rtint: the more i think about it, I think leaving it as is, is probably better for organization.

Assets/Script/UI/PlayerUserInterface.cs
m4rtint: no, cuz the sorting layer overrides everything.

Assets/Script/Managers/SoundEffectManager.cs
Gursaanj: for sure, we can look into more avenues when we get more sound effects and stuff

Gursaanj:
[Add SFX Manager for all scenes](https://trello.com/c/P7ZZxwfK/109-add-sfx-manager-for-all-scenes)

PR #48, opened 2019-08-23
Feature/enum player

m4rtint:
Branched from:
https://github.com/gotty2019/GOTTY/pull/44

Changed Player Number and strings into enum


Gursaanj:
[Use Enum instead of string "player 1"](https://trello.com/c/6yrWD3HY/106-use-enum-instead-of-string-player-1)

m4rtint:
[NPC based Levels wont load players in current Develop (PR 46)](https://trello.com/c/30vgfo00/111-npc-based-levels-wont-load-players-in-current-develop-pr-46)

Assets/Script/Managers/UIManager.cs
```diff
@@ -11,7 +19,7 @@ public class UIManager : MonoBehaviour
 
     private void Start()
     {
-        foreach(BasePanel panel in panels)
+        foreach (BasePanel panel in panels)
```
Gursaanj: Since its only a few panels a foreach is fine, but isnt a for loop more efficient if we are only doing one thing in the loop?

m4rtint: yeah, i just got lazy

PR #49, opened 2019-08-27
Feature/honk functionality

Gursaanj:
Cars can now Honk their horn (Only Keyboard Functionality at the moment (Using the q and right ctrl button))

The Sound Effect Manager is also better equipped for more sound effects at a time


m4rtint:
@spicyroboto I've been researching this for 2 weeks, there's no in built way of managing it. Either all in inspector, or there's some secret way that bigger companies do it and I'm not sure how.

PR #51, opened 2019-08-30
Set Pause and honk on controller

m4rtint:


m4rtint:
[Allow controller to select pause and end game](https://trello.com/c/zWVtKqkg/98-allow-controller-to-select-pause-and-end-game)

m4rtint:
[Allow controller to select pause and end game](https://trello.com/c/zWVtKqkg/98-allow-controller-to-select-pause-and-end-game)

Assets/Script/Managers/InputManager.cs
```diff
@@ -75,39 +75,34 @@ public static bool HandBreak()
         /// </summary>
         public static void PauseAndPlay()
         {
-            if (Input.GetButtonUp("Pause"))
+            if (Input.GetButtonDown("Pause") || IsStartPressed())
```
Gursaanj: I understand that were are using the start button on controllers for the pause button , so should we rename the bool to IsStartButtonPressed() instead? 

Could be a confusing naming convention later on if we give the controllers the ability to start the game from the main menu, and in turn, using a similar naming convention for a bool

m4rtint: IsStartButtonPressed - the prefix Is, usually means that it's returning a boolean.

PR #62, opened 2019-09-17
Enhancment/better feel

m4rtint:


Gursaanj:
[Create an initial Better feel to the game](https://trello.com/c/hO9I63lw/183-create-an-initial-better-feel-to-the-game)

Assets/Script/Game/Car/CarController.cs
```diff
@@ -81,11 +82,17 @@ private void OnDestroy()
...
     private void CarControllerOnUpdate()
     {
-        float vertical = GOTTY.InputManager.MainVertical(m_playerControllerInfo);
+        float vertical = AngryGG.InputManager.MainVertical(m_playerControllerInfo);
 
-        if (GOTTY.InputManager.IsBreaking(m_playerControllerInfo))
+        if (AngryGG.InputManager.IsBreaking(m_playerControllerInfo))
         {
-            vertical = 0;
+            m_breakFactor -= 0.001f;
```
wbeltran777: In the future let's make this 0.001f a variable that we can modify. We might find that it's more comfortable for the break to reduce the Forward Velocity more quickly.

m4rtint: This is basically how fast you break right?

PR #67, opened 2019-10-10
Feature/car selection

m4rtint:
IMPORTANT PART - InputManager


Assets/Script/Data/GameData.cs
```diff
@@ -12,10 +12,10 @@
 public static class GameData
 {
     private static int m_playerCount = -1;
-    private static List<PlayerInformation> m_playerInformationList = new List<PlayerInformation>()
+    private static Dictionary<Player, PlayerInformation> m_playerInformationList = new Dictionary<Player, PlayerInformation>
     {
-        new PlayerInformation(Player.Player_1, new PlayerDataLoader(), new PlayerControllerInformation(PCControl.WASD)),
-        new PlayerInformation(Player.Player_2, new PlayerDataLoader(), new PlayerControllerInformation(PCControl.KEYS))
+        {Player.Player_1, new PlayerInformation(Player.Player_1, 0, new PlayerDataLoader(), new PlayerControllerInformation(PCControl.WASD)) },
```
m4rtint: Changes to the playerinformation class

Assets/Script/Data/PlayerDataLoader.cs
```diff
@@ -72,6 +66,25 @@ public Color PlayerColor(Player player)
...
+        int amountOfCars = m_carSprites.Length;
+
+        this.m_AllCars = new CarObj[amountOfCars];
+        for (int i = 0; i < amountOfCars; i++)
+        {
+            m_AllCars[i] = new CarObj(m_carSprites[i]);
+        }
+    }
+
+    public struct CarObj {
```
m4rtint: I was planning for the player choose a CarObj - containing player's color and car sprite

BUT - problems happened, but leaving this here for now. 
Details in Utilities class

Assets/Script/Data/PlayerInformation.cs
```diff
@@ -27,13 +27,13 @@ public class PlayerInformation
...
         /// <param name="controllerInfo">Players controller input information</param>
-        public PlayerInformation(Player number, PlayerDataLoader loader, PlayerControllerInformation controllerInfo)
+        public PlayerInformation(Player number, int pointer, PlayerDataLoader loader, PlayerControllerInformation controllerInfo)
         {
             this.m_playerNumber = number;
             this.m_loader = loader;
             this.m_playerControllerInformation = controllerInfo;
             this.m_color = loader.PlayerColor(number);
-            this.m_sprite = loader.PlayerCarSprite(number);
+            this.m_sprite = loader.PlayerCarSprite(pointer);
```
m4rtint: loader.PlayerCarSprite => array of car sprites.

the pointer, just points to which index of car sprites

Assets/Script/Data/Utilities.cs
```diff
@@ -28,5 +28,50 @@ public static Color ColorFrom255(float r, float g, float b, float alpha = m_Full
         {
             return new Color(r / m_FullValue, g / m_FullValue, b / m_FullValue, alpha / m_FullValue);
         }
+
+        public static Color AverageColor(Texture2D texture)
```
m4rtint: Was planning to either get average sprite color OR dominant color.

BUT problems:
it takes the WHOLE car sprite sheet which contains every single car. 
I already cut up the car sprite sheet, and passed in a single car sprite, but the "texture" is still the WHOLE car sprite.

These functions are useless for now, but will be used in the future..maybe.

Assets/Script/Managers/InputManager.cs
```diff
@@ -16,6 +16,43 @@ namespace AngryGooseGames
     /// </summary>
     public static class InputManager
     {
+
```
m4rtint: I rearranged the methods around. 
Put regions in this.

AllTypes (checks all input method presses)
PC exclusive
Controller exclusive

Assets/Script/Managers/InputManager.cs
```diff
@@ -55,46 +92,92 @@ public static bool IsBreaking(PlayerControllerInformation inputInfo)
...
-        /// Pause or Resume the game
-        /// </summary>
-        public static void PauseAndPlay()
+        private static bool IsPCControlBreakPressed(PCControl control)
         {
-            if (Input.GetButtonDown("Pause") || IsStartPressed())
+            return Input.GetButtonDown(control + "_Break");
+        }
+
+        public static float IsPCControlHorizontalPressed(PCControl control)
```
m4rtint: @mikeslgoh @wbeltran777 @Gursaanj 
Off the top of my head, the only way I can figure this out. 

 When you do get axis, it slowly goes from 0 -> 1/-1.

So car selection would be called many many times. since it checks (isHorizontalPressed > 1)


This solution is to basically detect a button/key pressed ONCE and ONLY ONCE per frame


If you have any better solution please let me know

Assets/Script/UI/ControllerSelection.cs
```diff
@@ -60,45 +72,77 @@ private void AddControllersIfNeeded()
...
-        m_amountOfPlayersPlaying++;
+        m_playerInfoList.Add(player, new PlayerInformation(player, pointer, m_playerDataLoader, info));
         UpdateGameData();
+        UpdateReadyAmountOfPlayers(true);
+        return true;
+    }
+
+    private void UpdateReadyButtonIfNeeded()
+    {
+        m_nextButton.interactable = m_amountOfReadyPlayers == m_amountOfPlayersPlaying;
```
m4rtint: the "ready" button is only activated when everyone is ready

Assets/Script/UI/ControllerSelectionPanel.cs
```diff
@@ -3,33 +3,288 @@
...
+        m_border.color = m_loader.AllCars[m_currentPointer].m_Color;
+        m_carSprite.sprite = m_loader.AllCars[m_currentPointer].m_Sprite;
+    }
+
+    private int NumberOfCars()
+    {
+        return m_loader.AllCars.Length;
+    }
+
+    private void CheckDependencies()
```
m4rtint: Dependency checks.  let's see if it's easier to understand any broken builds.

Assets/Script/Data/Utilities.cs
mikeslgoh: Do you mean that you cut up the car sprite sheet outside of Unity and it still uses the WHOLE sheet with 4 cars?

m4rtint: Unity had a function.
Milticut sprite

Where you cut out a sprite sheet into individual sprites

m4rtint: If you go to Th resources folder under sprites, you'll get what I mean.

And yes it still takes the whole sprite with 9 cars

Assets/Script/Managers/InputManager.cs
mikeslgoh: I think once key press is the way to go. Do we have to put the check there? Because I thought for Key/Button down, it would only work when it is pressed and that's all

Assets/Script/UI/ControllerSelection.cs
mikeslgoh: So it only detects if it is waiting for a player if they are on the car selection right? It doesn't matter if a controller is plugged in as long as they don't join the game?

Assets/Script/UI/ControllerSelectionPanel.cs
mikeslgoh: Let's see how this works and then maybe consider adding it in other areas of our code base.

Assets/Script/Managers/InputManager.cs
m4rtint: Yeah exactly. I'm checking if it's pressed once. Not anymore.

Assets/Script/UI/ControllerSelection.cs
m4rtint: Exactly.

Assets/Script/Data/Utilities.cs
mikeslgoh: maybe we should consider trying to split it outside of Unity?

Assets/Script/Managers/InputManager.cs
mikeslgoh: ok yeah I think this is the best approach

Assets/Script/Data/Utilities.cs
m4rtint: I think that's the only way.

m4rtint: I'll probably do this afterwards

Assets/Script/UI/ControllerSelection.cs
```diff
@@ -40,11 +42,21 @@ public class ControllerSelection : MonoBehaviour
...
+    private void Start()
+    {
+        foreach(ControllerSelectionPanel panel in m_players)
+        {
+            panel.OnCarSelected += AddPlayerInfo;
+            panel.OnRemoveSelectedCar += RemovePlayerInfo;
+        }
+    }
+
     private void Update()
```
wbeltran777: Should we be subscribing this to the global update in world manager?

m4rtint: WorldManager is not in this menu scene yet.

And usually we have a worldmanager global update to pause/play/slow down time for stuff, but menu don't have that stuff.  - that's what i'm presuming

mikeslgoh: I think we have a task to do a UIDelegate I believe

m4rtint: If i remember correctly, the UI Delegate is more so to update the score on the game scenes

mikeslgoh: ah ok, that's probably it. Do we need a menu/panel update then? For consistency?

m4rtint: So for the worldmanager is good to do it so that you can pause/play/slow down time from ONE place only.

If there's a reason for the main menu, sure. But I can't think of a reason 

The main usage of the update in menu is to detect user input. 
Only reasoning I can think of, is to "disable user input".

mikeslgoh: oh sorry. Uhh I say that if there is no need to make this change except for just consistency, then it's fine to leave it as it is.

@Gursaanj @wbeltran777 ?

Assets/Script/UI/ControllerSelectionPanel.cs
```diff
@@ -3,33 +3,288 @@
...
+        {
+            ScrollThroughCars(true);
+        }
+        else if (m_inputDevice.LeftStickRight.WasPressed)
+        {
+            ScrollThroughCars(false);
+        }
+        else if (m_inputDevice.Action1.WasPressed)
+        {
+            CarSelected(m_inputDevice, m_playerNumber, m_currentPointer);
```
wbeltran777: Small bug, if a player presses 'down' as their method of connecting their controller, then they'll also select the first car on the sprite list as long as the car is not taken. This might cause slight confusion for showcasing as the first player to join may sometimes be set to READY right away and may be confused as to why they weren't able to pick a car.

Assets/Script/UI/ControllerSelection.cs
wbeltran777: Stuff seems to be working fine, so if we don't see any long term damage to not doing a global manager for UI, I say we don't worry about it.

m4rtint:
[Player Selection - Choose Cars](https://trello.com/c/Z0FJPPPq/212-player-selection-choose-cars)

Assets/Script/Managers/InputManager.cs
Gursaanj: Yup I thinks thats a good approach

Assets/Script/UI/ControllerSelection.cs
Gursaanj: Fair enough, Im down either way. but if this is the only case where update is begin called of the menu scene, then I guess we dont have to worry about it

Assets/Script/UI/ControllerSelectionPanel.cs
m4rtint: Lol that ain't a small bug. 

I think that part will be solved, when we figure out UX wise, how to "go next" .

Cuz I think we're thinking of pressing start to go next, instead of press "down" / "Accerlate"/"confirm"

Assets/Script/UI/ControllerSelection.cs
m4rtint: I think @wbeltran777 has a point. 

I'm thinking that we'll probably implement animated transitions between menu scenes.. And during the transitions, the players should not be able to press any buttons. 

High chance to have a global main menu delegate.

m4rtint: But that's for another PR

Assets/Script/UI/ControllerSelectionPanel.cs
wbeltran777: Okay gotcha! In that case things look pretty great!

m4rtint: Oh wait, i think i misunderstood what you meant. Now I understand. 

The InControl framework handles i nicely. it only detects the "OnButttonPressed" ONCE, so it won't detect it more than once.

wbeltran777: Right, but it's still triggering both, sync controller, and "Ready" at the same time, if you press 'S' or 
'Down Arrow' to join. Is this an issue we should be concerned about?

m4rtint: Yes, I'll fix it in this branch

m4rtint: @wbeltran777 
Fixed it, not sure if it's a work around or fix but..changed ControllerSelection to use 
LateUpdate, and ControllerSelectionPanel still uses Update.

m4rtint: yeah just googled, this is not a good fix. But if it's merged. it works lol. Will think of a better solution soon.

mikeslgoh:
I'll approve once we adjust the look to match the mocks (which seems to be pretty straight forward). Other than that, everything else looks good :)

m4rtint:
@mikeslgoh 
I kinda wanna approve this first..and then the looks in another PR. 

this PR is big enough as it is.

mikeslgoh:
[Change way for players to join the match](https://trello.com/c/QiQIcgzi/206-change-way-for-players-to-join-the-match)

mikeslgoh:
[Change way for players to join the match](https://trello.com/c/QiQIcgzi/206-change-way-for-players-to-join-the-match)

PR #68, opened 2019-10-10
Fixed Double points bug

wbeltran777:
I just added an extra condition for randomizing the parking spots that makes sure that we don't pick the same one over and over again. I'm thinking that this may have been a cause for the double points bug. 
Been playing for a bit and I haven't been able to replicate it, so I think that may have fixed it. 

In the future we can probably think about a randomizer that does 'fake' random picking so that we don't pick a parking spot right next to the current one, but we can probably leave that for another PR.

Do let me know if any of you run into the bug again and I'll get right to it.


m4rtint:
This is actually a pretty good theory

Assets/Script/Managers/ParkingManager.cs
```diff
@@ -79,7 +79,8 @@ public void Initialize()
     /// </summary>
     public void ActivateRandomParkingSpot()
     {
-        GameObject activatedParkingSpot = m_targets.Where(i => i.activeSelf).OrderBy(n => Random.value).FirstOrDefault();
+        
+        GameObject activatedParkingSpot = m_targets.Where(i => i.activeSelf && i != CurrentTarget).OrderBy(n => Random.value).FirstOrDefault();
```
m4rtint: I swear @spicyroboto had a condition for this....strange.

Gursaanj:
[Players get two points for one parking spot gotten](https://trello.com/c/k5poAJ7G/217-players-get-two-points-for-one-parking-spot-gotten)

PR #69, opened 2019-10-11
Feature/menu screen scroll

wbeltran777:
Yo yo, I did a quick implementation of the background scrolling. Let me know what you guys think!


m4rtint:
Conflicts

Gursaanj:
[Make the car background move](https://trello.com/c/vedAgHHs/193-make-the-car-background-move)

Gursaanj:
Seems to loop through some empty background too!!, but other than that I think the speed is nice and smooth

Gotty broken screen slide

Assets/Script/UI/BackgroundBehaviour.cs
```diff
@@ -0,0 +1,43 @@
...
+/// <summary>
+/// In charge of behaviour for scrolling background in title screen
+/// </summary>
+public class BackgroundBehaviour : MonoBehaviour
+{
+
+    [SerializeField] private float m_scrollSpeed = 2.0f;
+    private RectTransform m_backgroundCenter;
+    private Vector3 m_originalPosition;
+    private Image m_backgroundImage;
```
Gursaanj: seems like m_backgroundImage is not being used

wbeltran777: Deleted it, also lol I promise I ran this before submitting. I'll look into why the image isn't scrolling correctly.

m4rtint: check up on different resolutions.

m4rtint: U might need to get the physical screen resolution, and then scroll according to the width.

wbeltran777:
Fixed the scene conflicts, the new scrolling background is in a prefab called MainMenuBackground in the UI prefab folder. 
To try out the scrolling just drag the prefab to replace 'Image' in the Main Menu Scene.
Main Camera -> Canvas -> Image

wbeltran777:
New prefab to drag in is Assets -> Prefab -> UI -> BackGroundScroll

Replace image in the main camera

Last implementation I used two images which did some weird stuff for distance when changing aspect ratios. 
This time it's just a long strip. 
The looping is a little bit janky, but I made the scrolling image long enough and the speed slow enough that it won't really be visible often. 
I know it's not an amazing fix, but I'd like to get something done before Saturday

PR #70, opened 2019-10-12
Uninstall git lfs

mikeslgoh:
I was able to push and create a PR for this branch so must have done something right. Had to try and remove AStar for now so will continue to see how this goes


mikeslgoh:
It takes awhile to push files compared to before but it's not bad. Just need to make smaller commits now.

mikeslgoh:
[Removed Git LFS from project](https://trello.com/c/MyGz9DMi/223-removed-git-lfs-from-project)

PR #72, opened 2019-10-14
Added a +1 animation whenever a player scores a point

mikeslgoh:
Used LeanTween to display a +1 animation whenever a player is awarded a point. Not sure whether I should have added it into UpdateScore since it would play at the same time or keep it as a separate function


Assets/Script/UI/PlayerUserInterface.cs
```diff
@@ -15,17 +15,32 @@
...
     public void UpdateScore(int score)
     {
-        displayScore.text = score.ToString();
+        m_displayScore.text = score.ToString();
+    }
+
+    /// <summary>
+    /// Plays the scored animation.
+    /// </summary>
+    public void PlayScoredAnimation()
```
mikeslgoh: This is the main code that I used to play the animation. Used the same code that is used to display "Already Taken" message on the main menu

mikeslgoh:
[DIsplay +1 when a player wins a point](https://trello.com/c/OYPkvlqR/222-display-1-when-a-player-wins-a-point)

m4rtint:
i don't know if it's just me, but naming the png +1, having the "+" sign gives me the heebee jeebes. 
Like some string error/warning would popup.

mikeslgoh:
> i don't know if it's just me, but naming the png +1, having the "+" sign gives me the heebee jeebes.
> Like some string error/warning would popup.

I can rename it to scoredOne?

Assets/Script/UI/PlayerUserInterface.cs
```diff
@@ -15,17 +15,32 @@
...
 {
     [SerializeField]
-    private TextMeshProUGUI displayScore;
+    private TextMeshProUGUI m_displayScore;
     [SerializeField]
-    private Image border;
+    private Image m_border;
+    [SerializeField]
+    private GameObject m_scored;
+    private const float m_scoredAnimationTime = 0.25f;
```
Gursaanj: maybe change this to serialize field first so that we can more easily play around with this value to get a better feel (i.e. when we have our playtest session)

Assets/Script/UI/PlayerUserInterface.cs
```diff
@@ -15,17 +15,32 @@
...
+    /// <summary>
+    /// Plays the scored animation.
+    /// </summary>
+    public void PlayScoredAnimation()
+    {
+        m_scored.LeanScale(Vector3.one, m_scoredAnimationTime)
+           .setEaseOutBack();
+        m_scored.LeanScale(Vector3.zero, m_scoredAnimationTime)
+            .setEaseInBack()
+            .setDelay(m_scoredAnimationTime * 3);
```
Gursaanj: why is there a *3 ?

mikeslgoh: This was based on a value set in MainMenu for the already taken text. I can set this as a variable so that we can play around with this as well?

Assets/Script/UI/PlayerUserInterface.cs
mikeslgoh: I just checked and this script does not appear in the hierarchy so we would still need to change it through the code

Assets/Script/UI/PlayerUserInterface.cs
m4rtint: so the * 3 is like
Step 1)
Pop Open - takes x seconds to do it
Step 2)
delay it by x seconds * 3....so that it shows for that extra time before fading out

m4rtint: yeah could be better explained. need to place this in some animation class next time

mikeslgoh: We can add it as part of the refactoring tasks?

m4rtint: yup

Gursaanj: Yup that sounds a good idea

PR #75, opened 2019-10-14
Feature/setPlayerBorderColor

mikeslgoh:
Border colours are now set based on the chosen car. Separated the car sprite sheet into individual cars (removed some cars that are too similar to one another).


Assets/Script/Data/PlayerDataLoader.cs
```diff
@@ -43,21 +43,10 @@ public Sprite PlayerCarSprite(int carNumber)
...
-            case Player.Player_2:
-                return Utilities.ColorFrom255(255, 215, 14);
-            case Player.Player_3:
-                return Utilities.ColorFrom255(103, 173, 31);
-            case Player.Player_4:
-                return Utilities.ColorFrom255(81, 159, 220);
-            default:
-                return Utilities.ColorFrom255(99, 98, 93);
-        }
+        int index = Mathf.Clamp(carNumber, 0, AllCars.Length - 1);
```
mikeslgoh: Since I was able to set the color based on the sprite, I did not need to use the Utilities functions.

Assets/Script/Data/PlayerDataLoader.cs
```diff
@@ -83,8 +72,11 @@ public struct CarObj {
         public CarObj(Sprite sprite)
         {
             this.m_Sprite = sprite;
-            //TODO - need to get color from car at some point. Need to import seperate sprites
-            this.m_Color = Color.white;
+            // Get the sprite colour from the front of the car hood
+            Rect rect = sprite.rect;
+            int x = Mathf.FloorToInt(rect.width / 2);
+            int y = Mathf.FloorToInt(rect.height - (rect.height / 6));
+            this.m_Color = sprite.texture.GetPixel(x,y);
```
mikeslgoh: When I got the colour, the RGB scale was not based on 255 but instead 0-1. Should I try and figure out a way to get to use the 255 scale?

mikeslgoh:
[Change the border colour to match the chosen car](https://trello.com/c/rak9gwlC/226-change-the-border-colour-to-match-the-chosen-car)

mikeslgoh:
[Change the border colour to match the chosen car](https://trello.com/c/rak9gwlC/226-change-the-border-colour-to-match-the-chosen-car)

Assets/Script/Data/PlayerDataLoader.cs
m4rtint: I'm going to take a look at average color, and then try to fix dominant color on your branch, see if they look good.

mikeslgoh: ok cool but so far, implementation is fine?

m4rtint: yeah

Gursaanj: hmm if you could just multiply the number by 255 to convert it to the 255 RGB scale if you think that would be better?

mikeslgoh: Martin's looking into it now to see whether there's a better way to get the color rather than an arbitrary pixel.

PR #76, opened 2019-10-14
Refactor/input manager

m4rtint:
Refactored inputmanager

Seperated controller from PC


m4rtint:
I haven't tested it with a controller yet, don't have access to one.

So if anyone wants to test it

Assets/Script/Managers/Input/ControllerInputManager.cs
```diff
@@ -0,0 +1,82 @@
...
+
+            public static bool IsCancelPressed(InputDevice ip = null)
+            {
+                InputDevice device = ip ?? InControl.InputManager.ActiveDevice;
+                return device.Action2.WasPressed;
+            }
+            #endregion
+
+            #region GamePlayer
+            public static bool IsBreakingPressed(InputDevice ip)
```
mikeslgoh: Can we rename this to either isHandbrakePressed since this is related to the handbrake button?

Assets/Script/Managers/Input/InputManager.cs
```diff
@@ -0,0 +1,126 @@
...
+
+        #endregion
+
+
+        private static bool IsInputPC(PlayerControllerInformation inputInfo)
+        {
+            return inputInfo.ControllerPlatform == ControllerPlatform.PC;
+        }
+
+        private static float VerticalController(InputDevice device)
```
mikeslgoh: Is there a reason why we didn't move this into ControllerInputManager?

Assets/Script/Managers/Input/ControllerInputManager.cs
m4rtint: sure. 
So there is parkingbreaking and handbreak?

mikeslgoh: well our parkingbreak is our handbreak though or that's what we're planning to test at playtesting

Assets/Script/Managers/Input/InputManager.cs
m4rtint: Oh...yeah i'll move it

mikeslgoh:
Your InputManager works the same as develop although I found a few bugs in develop

m4rtint:
I removed the TODO from map selector

mikeslgoh:
[InputManager Refactor](https://trello.com/c/Cy2JBD80/225-inputmanager-refactor)

PR #77, opened 2019-10-14
refactor/use dominant color for broder

mikeslgoh:
Use the dominant color of a sprite texture rather than an arbitrary point to set the border color for a player

Question: color vs colour?


m4rtint:
Whatever unity uses. That's the answer.

PR #79, opened 2019-10-14
Changes the control scheme animation

m4rtint:
https://imgur.com/FI4XC7z

Made it more in tune with the mock ups


Assets/Scenes/MainMenu.unity
```diff
@@ -517,6 +517,11 @@ PrefabInstance:
       propertyPath: m_inputSource
```
m4rtint: Changes in main menu is for the quit panel

PR #80, opened 2019-10-15
Enhancement/back button animation

mikeslgoh:
Added back button animation and grabbed the back button arrow and text from Figma

The prefab includes 2 arrows: the base and then an arrow with the loading colour.
When the back button is pressed, the loading arrow will fill from right to left.


Assets/Script/Managers/MainMenuManager.cs
```diff
@@ -92,6 +95,18 @@ private void OnBackward()
             DisableScreen(m_states);
             m_states--;
             UpdateView();
+            ShowBackButton();
+        }
+    }
+
+    private void ShowBackButton()
```
mikeslgoh: This ensures that the back button is disabled when on the START screen but visible on the other screens

Assets/Script/UI/MainMenu/BackButton.cs
```diff
@@ -37,6 +40,19 @@ private bool IsBackButtonHeldDown()
         return isPCBackButtonHeldDown || isControllerBackButtonHeldDown;
     }
 
+    private void updateLoadingButton()
+    {
+        if (!float.IsPositiveInfinity(m_currentTimer))
```
mikeslgoh: Need this check because I was running into an issue where the image was loading even though the back button functions were not being called

Assets/Script/UI/MainMenu/BackButton.cs
```diff
@@ -46,12 +62,16 @@ private void Update()
...
+            updateLoadingButton();
 
             if (m_currentTimer <= 0)
             {
                 m_currentTimer = float.PositiveInfinity;
                 OnBackButtonHeld();
+                resetLoadingButton();
             }
+        } else {
+            resetLoadingButton();
```
mikeslgoh: Not sure if this was the best way to do it but whenever a player lets go of the back button, it reset to the base arrow

mikeslgoh:
[Create the back button animation](https://trello.com/c/lICRS63I/229-create-the-back-button-animation)

Assets/Script/UI/MainMenu/BackButton.cs
```diff
@@ -6,13 +6,16 @@
...
+using UnityEngine.UI;
 using System;
 
 public class BackButton : MonoBehaviour
 {
     public event Action OnBackButtonHeld;
 
     [SerializeField] private float m_durationToHold = 3.0f;
+    [SerializeField] Image loadingBackButton;
+    [SerializeField] float loadingTimeConstant = 0.008f;
```
Gursaanj: hmm this number should be based on m_durationToHold so that if ever change the duration for the player to press the back button, loadingTimeConstant can be updated automatically

Assets/Script/UI/MainMenu/BackButton.cs
Gursaanj: Should be fine for now I think. we could always create a IsBackButtonBeingReleased bool for this case too, if this doesnt seem right

mikeslgoh: Yeah I'm trying now for button released and looks a lot better to do it that way

m4rtint: Everything looks good.

PR #81, opened 2019-10-15
bugfix/select map with controllers

mikeslgoh:
This might be a temp fix but I made so that the map that is being previewed is currently selected. This allows a controller to press X/A and start the game.


Gursaanj:
[Unable to select map to start game with a controller](https://trello.com/c/zoofcz99/233-unable-to-select-map-to-start-game-with-a-controller)

PR #82, opened 2019-10-15
Enhancment/indicate back button user

Gursaanj:
The text next to the back button will change depending on what controller the player is using. I.e. if using the keyboard, players pressing the ESC key will see "Hold the Esc Key" next to the back button. Playstation Users will see "Hold the O Button". Xbox and Unidentified users will see "Hold the B Button".

CANNOT MERGE UNTIL PR #80 is merged
BackButtonText


Gursaanj:
If Someone Could try this with the XBox controller too that would be nice

Gursaanj:
[Display which button is being held depending on who is holding the back button](https://trello.com/c/WPfcIf6k/228-display-which-button-is-being-held-depending-on-who-is-holding-the-back-button)

Assets/Script/UI/MainMenu/BackButton.cs
```diff
@@ -57,6 +79,33 @@ private void updateLoadingButton()
...
+    private void GetPCPromptForBackButton()
+    {
+        m_playerUsingBackButton.text = m_initialBackButtonText + m_pcPrompt;
+    }
+
+    private void GetControllerPromptForBackButton(string controllerName)
+    {
+        string buttonPromptText;
+
+        if (controllerName.ToLower().Contains("playstation"))
```
m4rtint: This shouldn't check if it's playstation of xbox.

1) it'll never be PC.
2)It'll show PC if you plugin some china made controller

So you should just do simply - 
buttonPromptText = controllerName;

Assets/Script/UI/MainMenu/BackButton.cs
```diff
@@ -57,6 +79,33 @@ private void updateLoadingButton()
...
+        {
+            buttonPromptText = m_playstationPrompt;
+        }
+
```
m4rtint: remove extra space for this else

Assets/Script/UI/MainMenu/BackButton.cs
mikeslgoh: But how will we be able to decide between the xbox or playstation prompt? Unless there's a way to just get the button being pressed

m4rtint: controller Name should already contain - PlayStation. 
if it's too long and it say "SOny Playstation" 

then the else statement could be the just 
buttonPromptText = controllerName

mikeslgoh: but for the prompt we want it to say **HOLD O button** or **HOLD B button** so how can we set that my just setting buttonPromptText = controller name?

m4rtint: Oh wait shit, now i understand what this does. Nevermind

mikeslgoh: I do agree with removing the PC check in this function since it shouldn't reach this point

Assets/Script/UI/MainMenu/BackButton.cs
m4rtint: I'll approve this, but if u could change this before the merge, that'll be nice

PR #84, opened 2019-10-16
Bugfix/minor fix part3

mikeslgoh:
**Waiting for Martin's Input Manager refactor.**

- Removed check for the Cancel Button
- The highlight disappears between START and OPTIONS because it jumps to the QuitMenu that is still active. I made some changes to the quit panel but not sure if this is what I should be doing


mikeslgoh:
[Highlight disappears when going between PLAY and OPTIONS button](https://trello.com/c/mwyX08hX/81-highlight-disappears-when-going-between-play-and-options-button)

mikeslgoh:
[Back button allows players to join game](https://trello.com/c/UDzJvdmj/232-back-button-allows-players-to-join-game)

Assets/Script/UI/MainMenu/QuitPanel.cs
```diff
@@ -49,9 +58,18 @@ private void Awake()
         m_no.onClick.AddListener(() =>
         {
             OpenClosePanel(false);
+            HideOrShow(this.gameObject, false);
```
m4rtint: It's not ideal..but it works.

mikeslgoh: We can go through and note down what we need to refactor after this Sat?

Assets/Script/Managers/Input/ControllerInputManager.cs
```diff
@@ -23,7 +23,7 @@ public static bool IsStartPressed(InputDevice ip = null)
 
             public static bool IsJoiningPressed(InputDevice ip)
             {
-                return IsConfirmPressed(ip) || IsCancelPressed(ip) || ip.Action3.WasPressed || ip.Action4.WasPressed;
+                return IsConfirmPressed(ip) || ip.Action3.WasPressed || ip.Action4.WasPressed;
```
m4rtint: I think this should stay. and then fix the ready up button

m4rtint: cuz then u're basically creating a another bug. we have sprites that say  "press these 4 buttons"

mikeslgoh: I guess we can playtest and see whether it's actually an issue or not

Assets/Script/Managers/Input/ControllerInputManager.cs
```diff
@@ -55,6 +55,12 @@ public static bool IsCancelReleased(InputDevice ip = null)
                 InputDevice device = ip ?? InControl.InputManager.ActiveDevice;
                 return device.Action2.WasReleased;
             }
+
+            public static bool IsCancelHeld(InputDevice ip = null)
+            {
+                InputDevice device = ip ?? InControl.InputManager.ActiveDevice;
+                return device.Action2.IsPressed;
+            }
```
Gursaanj: Im doing this just in case we ever want to change the reverse button in game , we will not effect the cancelling button for the Menu stuff.. If that makes sense

Assets/Script/Managers/Input/ControllerInputManager.cs
mikeslgoh: @Gursaanj if you're able to, can you reverse this change as well?

Assets/Script/Managers/Input/ControllerInputManager.cs
Gursaanj: cause before this commit we had it so that we were checking if the car was reversing to check if the cancel button is pressed (same button)

m4rtint: just return IsReversing(ip)
instead of all that

Gursaanj: hmm but the whole point of this was so that if we change what the reverse button was i.e changing it from O to like Left Trigger. Then we dont have to remember to change the cancelling button too

mikeslgoh: Another thought, would be that if we decide to support both O and Left Trigger for reversing, then Left Trigger could be used as cancelling if we use isReversing() so might be good to use this?

Gursaanj: Yeah thats what Im thinking

m4rtint: Sure.

PR #88, opened 2019-10-17
Enhancement/sfx management udd

Gursaanj:
Created a system where multiple SFX can be alternately be played for certain events. and added a second SFX for Obstacle collision. Also lowered the volumes of Player and Obstacle Collision SFXs (hardcoded) for UDD so people can hear and evaluate the background music better.


Gursaanj:
[Change/eliminate obstacle collision noises or turn down frequency that the audio clip is played](https://trello.com/c/fpoIhGoe/216-change-eliminate-obstacle-collision-noises-or-turn-down-frequency-that-the-audio-clip-is-played)

Assets/Scenes/Gazebo.unity
```diff
@@ -439,6 +439,16 @@ PrefabInstance:
       propertyPath: m_LocalEulerAnglesHint.z
       value: 0
       objectReference: {fileID: 0}
+    - target: {fileID: 5290987832730758441, guid: 42e516a9f57b148dcb1eba5649f9f4eb,
+        type: 3}
+      propertyPath: m_parkedSFX.Array.size
```
mikeslgoh: did you add a test prefab in Gazebo?

Assets/Script/Game/Car/CarController.cs
```diff
@@ -181,15 +181,16 @@ private void OnCollisionEnter2D(Collision2D collision)
     {
         float m_scaledVolume = (collision.relativeVelocity.magnitude - m_MinRelativeVelocity) / (m_MaxRelativeVelocity - m_MinRelativeVelocity);
 
-        // To make sure all volume is scaled between 0 and 1
-        m_scaledVolume = Mathf.Clamp(m_scaledVolume, 0, 1);
+        // To make sure all volume is scaled between 0 and 1 
+        // Note: Literally just Halved the volume so that Background music is audible. Will be adressed after UDD *TODO
```
mikeslgoh: Normally for this you can just do - TODO: address ...
This is fine for now but just letting you know :)

Assets/Scenes/Gazebo.unity
Gursaanj: Nah so I added the parking SFX to the Gazaebo SoundEffect Manager accidentally instead of adding it to the sound effect manager prefab, then added the same SFX to the prefab. So I think it is just double counting if that makes sense

Gursaanj:
> Just left a couple of questions. Also, did you add new SFX?

Yeah Obstacle Collision Ding II but thats all. Is it not audible? or should I try to find more?

Assets/Script/Managers/SoundEffectManager.cs
```diff
@@ -45,23 +45,29 @@ public class SoundEffectManager : MonoBehaviour
...
+    //private Queue<AudioSource> m_audioSources;
+
+    private AudioSource m_audioSource;
 
     //// Create seperations in the serialized fields so that placing more audio clips will be more organized 
 
     [Header("Car SFX")]
-    [SerializeField] private AudioClip m_carHornSFX;
+    [SerializeField] private AudioClip[] m_carHornSFX;
+    private int m_carHormSFXIndex = 0;
```
m4rtint: I'm sure there's a data structure you can use to keep track of all these indexes. instead of a new one for every single type of sound.

mikeslgoh: We can create a Trello card to keep track then so that we look into it after UDD?

PR #89, opened 2019-10-22
Enhancement/dpad movement recognition

Gursaanj:
Allows users with controllers to use the Dpad to select players and Maps in addition to the left stick


Gursaanj:
[Allow Controllers use Dpad to select Cars and Maps](https://trello.com/c/u1rwcnmC/254-allow-controllers-use-dpad-to-select-cars-and-maps)

PR #91, opened 2019-10-23
Refactored Common animation into one script

m4rtint:


m4rtint:
[Create an Animation class](https://trello.com/c/zKcvuCbN/227-create-an-animation-class)

Assets/Script/UI/GameCountDown.cs
```diff
@@ -12,8 +12,10 @@
 /// <summary>
 /// On Start of game do a count down before starting the game
 /// </summary>
+[RequireComponent(typeof(TextMeshProUGUI))]
 public class GameCountDown : MonoBehaviour
 {
+    internal readonly Vector3 m_EnlargedCountDownScale = Vector3.one * 3;
```
mikeslgoh: what does an **interal readonly** modifier do differently than a **private const**?

Assets/Script/Data/CommonAnimations.cs
```diff
@@ -0,0 +1,28 @@
+﻿using UnityEngine;
+
+public static class CommonAnimations
```
mikeslgoh: @wbeltran777 might want to take a look at this since you mentioned you might be doing some animation work for EndGamePanel?

Assets/Script/UI/GameCountDown.cs
m4rtint: internal is the same as public, except that it is only visible inside the assembly it is delcared in.  (SDK/API cases). 

Private is only visible inside the declaring type

Gursaanj:
Yeah I agree with @mikeslgoh , Lets not merge the PR until the end game panel is done.

m4rtint:
@Gursaanj How come after end game panel is done?

m4rtint:
@Gursaanj shouldn't this be merged in so that End game panel can utilize these functions?

Gursaanj:
@m4rtint  hmm it depends if Will has already started on the End game panel I guess?? Cause if He has already started then might not want him to refactor as well. Or if he is down, I guess he can use this ?? Up to you guys ??

m4rtint:
@Gursaanj he can do the end panel without this refactored animation part anyway.

mikeslgoh:
I meant that I wanted Will to review this PR first and then merge. We don't have to wait till the EndGamePanel is merged

PR #93, opened 2019-10-28
Refactor/score manager

m4rtint:
- Removed it from being a singleton
- Refactored a few other things.

It should run, no feature changes


Assets/Prefabs/Managers/Managers.prefab
```diff
@@ -47,19 +46,6 @@ MonoBehaviour:
   m_Name: 
   m_EditorClassIdentifier: 
   m_timeTakenToCompleteParking: 0.5
---- !u!114 &2817640143745387221
```
m4rtint: Removed scoremanager from the manager prefab

Assets/Script/Game/PlayerController.cs
```diff
@@ -47,13 +47,28 @@ public int PlayerNumber()
...
         CreateCamera();
         SetupUserInterface(playerInfo.Color); 
     }
 
+    private void OnDestroy()
+    {
+        ScoreManager.OnPlayerScored -= UpdateScoreUI;
+    }
+
+    private void UpdateScoreUI(Player playerScored, int score)
```
m4rtint: I call this whenever there is an update on Score. 
If ANYONE scores a point, all the players get an update.

So check if it's the correct player, then update.

m4rtint:
[ScoreManager refactor](https://trello.com/c/1Fh8fvuV/121-scoremanager-refactor)

Assets/Script/Managers/ParkingManager.cs
```diff
@@ -79,17 +73,13 @@ public void Initialize()
     /// </summary>
     public void ActivateRandomParkingSpot()
     {
-        
-        GameObject activatedParkingSpot = m_targets.Where(i => i.activeSelf && i != CurrentTarget).OrderBy(n => Random.value).FirstOrDefault();
+        GameObject activatedParkingSpot = m_targets.Where(i => i.activeSelf && i != CurrentTarget).OrderBy(n => UnityEngine.Random.value).FirstOrDefault();
```
mikeslgoh: What is the difference with using **Random.value** and **UnityEngine.Random.value**?

m4rtint: uhhhhh one is that Unity does it, the other is that C# does it. Doesn't really matter with our context

Assets/Script/Managers/WorldManagers/BaseWorldManager.cs
```diff
@@ -37,35 +34,29 @@ public abstract class BaseWorldManager : MonoBehaviour
...
-        {
-            m_scoreManager = GetComponent<ScoreManager>();
-        }
-
         if (m_playerManager == null)
         {
             m_playerManager = GetComponent<PlayerManager>();
         }
+
+        m_parkingManager.OnParkingChange += SetNewParkingSpot;
```
mikeslgoh: just wondering why we moved this to Awake() instead of leaving it in Start()?

m4rtint: personally i put "getcomponent" or like initializing variables/methods in awake
- The life cycle is Awake -> OnEnable -> Reset -> Start
So that delegates are attached before anything else.

PR #95, opened 2019-10-29
Forward  Button, made backbutton with abstract

m4rtint:
Almost Done.
If u want to look through the code for now.


Assets/Script/UI/MainMenu/HoldButton/HoldButton.cs
```diff
@@ -0,0 +1,119 @@
+﻿using System;
+using UnityEngine;
+using UnityEngine.UI;
+using TMPro;
+using InControl;
+
+public abstract class HoldButton : MonoBehaviour
```
m4rtint: All Logic is in Abstract class. 

only need to replace some text and which button needs pressed

m4rtint:
[EndGamePanel animation and button](https://trello.com/c/cSbJNvbb/259-endgamepanel-animation-and-button)

Assets/Script/Managers/WorldManagers/BaseWorldManager.cs
```diff
@@ -49,8 +49,8 @@ private void Awake()
 
     private void Start()
     {
-        m_parkingManager.Initialize();
         m_playerManager.Initialize();
+        m_parkingManager.Initialize();
```
m4rtint: Player needs to be made, Then the parking spot is created and assigned to the player

ORDER MATTERS - DEPENDENCIES

m4rtint:
[EndGamePanel animation and button](https://trello.com/c/cSbJNvbb/259-endgamepanel-animation-and-button)

Assets/Script/UI/MainMenu/HoldButton/HoldButton.cs
```diff
@@ -0,0 +1,126 @@
...
+    protected abstract string PlayStationButtonPrompt();
+
+    protected abstract string PCButtonPrompt();
+    
+    private string InitialButtonPrompt()
+    {
+        return UI.MainText.InitialHoldButtonMessage;
+    }
+
+    private void GetPCPromptForBackButton()
```
wbeltran777: If we're abstracting, should these methods that talk about the Back Button be kept in the Back Button class?

m4rtint: I misnamed it. 
Should be called GetPCPromptForButton()

m4rtint: Fixed it

PR #98, opened 2019-10-31
Addition/new audio manager

Gursaanj:
Imported a new Audio Manager (Papae Audio Manager: https://github.com/oluwaseyeayinla/Papae.AudioManager) and channeled all sound effects and Background music (not including the one in the Main Menu) through the new system 

This system inherently uses a pooling system for better audioclip usage and has a good amount of flexibility for using AudioMixer and AudioMixerGroups (which is the next step for better and smoother audio control)

If this system works with you guys, Ill start working on the AudioMixer Stuff as well as integrating the music/sfx of the main menu into the Sound Manager


Assets/Script/Managers/Input/InputManager.cs
```diff
@@ -137,7 +137,8 @@ private static void PauseAndResume()
             if (StateManager.Instance.IsPlaying() || StateManager.Instance.IsPaused())
             {
                 StateManager.Instance.PlayPauseSwitch();
-                SoundEffectManager.Instance.PlaySoundEffect(SOUNDEFFECT.PauseAndResume);
+                //SoundEffectManager.Instance.PlaySoundEffect(SOUNDEFFECT.PauseAndResume);
```
m4rtint: Remove this

Assets/Script/Managers/SoundManager.cs
```diff
@@ -0,0 +1,106 @@
...
+        }
+
+        AudioManager.Instance.PlayOneShot(sfx, transform.position, relativeVolume);
+    }
+
+    ///Event called when a new scene is loaded to play the appropriate Background music
+    ///Needs parameters to be used, no use for current parameters
+    private void PlayBackgroundMusic(Scene scene, LoadSceneMode mode)
+    {
+        AudioManager.Instance.PlayBGM(m_backgroundMusic);
```
m4rtint: None of the arguments here are used.

Assets/Script/Managers/SoundManager.cs
```diff
@@ -0,0 +1,106 @@
...
+        else
+        {
+            Destroy(gameObject);
+        }
+        DontDestroyOnLoad(gameObject);
+    }
+
+    private void OnEnable()
+    {
+        SceneManager.sceneLoaded += PlayBackgroundMusic;
```
m4rtint: Why on enable and disable?  why not onStart and onDestroy?

Assets/Script/Managers/SoundManager.cs
```diff
@@ -0,0 +1,106 @@
...
+using System.Collections.Generic;
+using UnityEngine;
+using UnityEngine.SceneManagement;
+using Papae.UnitySDK.Managers;
+
+public class SoundManager : MonoBehaviour
+{
+    private static SoundManager m_instance;
+
+    [Header("Background Music")]
```
m4rtint: If you want to, you can create a scriptableobject that stores all the audio.

Assets/Script/Managers/SoundManager.cs
Gursaanj: Yeah could try that out just followed what I remembered

Assets/Script/Managers/SoundManager.cs
Gursaanj: So Scenemanager.sceneloaded is a delegate that subsrcibes to events (methods) that have those parameters, I was looking around and couldnt find another way to signal between scene changes

Assets/Script/Managers/SoundManager.cs
Gursaanj: Yeah I might just do that. Currently having the issue where you replay a map or go back to the main menu and play the same map the Background music will stop playing

Gursaanj: Will try the scriptable objects instead seems better

Assets/Script/Managers/SoundManager.cs
m4rtint: I'm not sure if OnDisabled will be called when you change scenes. 
two options
1) Check if OnDisabled is called when changing scenes, if both enable and disable are being called when changing scenes, you're safe. 

2) Put it into ondestroy.

Assets/Script/Managers/SoundManager.cs
m4rtint: Oh i guess the delegate just needs to have those arguments. If it's a must, then you can just leave it.

Assets/Script/Managers/SoundManager.cs
m4rtint: it won't fix that problem, but it will look more organized.

Assets/Script/Managers/SoundManager.cs
Gursaanj: I did test with OnDisable (with Debug.log) and it seems to work, but I still think that On destroy might be better. Will try it out

Assets/Script/Managers/SoundManager.cs
```diff
@@ -0,0 +1,123 @@
...
+    /// The parameters are a byproduct of the inherent Scenemangement Event used to check if a scene has been loaded
+    /// </summary>
+    /// <param name="scene"></param>
+    /// <param name="mode"></param>
+    private void PlayBackgroundMusic(Scene scene, LoadSceneMode mode)
+    {
+        // If there is a mismatch between the list of audioclips and the number of scenes just play the MainMenuMusic
+        int backgroundMusicIndex = CheckForCorrectNumberOfScenes() ? scene.buildIndex : 0;
+        AudioManager.Instance.PlayBGM(m_backgroundMusic[backgroundMusicIndex]);
+    }
```
Gursaanj: So after discussing with Martin, there is currently no "clean" way that we could think of to link the scene to the appropriate Background Music so for now, I am using a bit of a hacky way in which I ensure that the array of Background music matches the list of Scenes in the Build settings. This will work for now (while Martin is looking into an asset to allow for serialized dictionaries, Easily cleaning up the situation)

PR #101, opened 2019-11-02
Enhancement/refactored audio manager

Gursaanj:
Imported an assets  that lets us Serialize dictionaries. Used it for easier to manage SoundEffects and Background Music Placement (since list would be too messy and harder to handle).

I didnt delete the example folder or the documentation just yet so that everyone else can take a look and get familiar with how it works. (Pretty intuitive)

Also refactored sceneinformation cause we were using the terms for scenes to describe map choices rather than the scenes actually used in build (had to to make Background music easier to handle for).

Made sure the the keys for the dictionaries were always based on Enums since then we dont have to use strings or int (where typos can occur)


Gursaanj:
Scripts to look at:

SoundManager.cs
SerializedDictionary.cs
SceneInformation.cs
Warnings.cs

the rest were from either renamings or the addition of the assets

Gursaanj:
[Import Seriliazble Dictionary Asset](https://trello.com/c/TnAMnyiD/275-import-seriliazble-dictionary-asset)

Gursaanj:
[Refactor Sound Manager with Serializable Dictionaries](https://trello.com/c/HXytEjAj/276-refactor-sound-manager-with-serializable-dictionaries)

Assets/Script/Data/SceneInformation.cs
```diff
@@ -5,10 +5,39 @@
...
+    /// The Office Map Scene
+    /// </summary>
+    Office
+}
+
+/// <summary>
+/// The Map names for Users to See 
+/// </summary>
+[System.Serializable]
+public enum MAPNAME
```
m4rtint: Don't have 2 different enums that have the same values. 
it's super confusing. 
stick with SCENENAME or MAPNAME, combine both into one. 

I would assume
MainMenu, Fountain, Gezbo,office, random

And whenever you have to handle the enum, just do checks and balances.

Assets/Script/UI/MainMenu/MapSelectionManager.cs
```diff
@@ -98,8 +98,8 @@ private void AnimateMapScale(int currentPointer, bool isLeftPressed, float timeT
     /// </summary>
     public void Initialize()
     {
-        SCENENAME[] listOfSceneNames = SceneInformation.ListOfSceneNames;
-        foreach (SCENENAME level in listOfSceneNames)
+        MAPNAME[] listOfSceneNames = SceneInformation.ListOfSceneNames;
+        foreach (MAPNAME level in listOfSceneNames)
```
m4rtint: Assuming you combine all the enums together. 

here you can do a check


if (level != MAINMENU){
      doStuff();
}

Assets/Script/Data/SceneInformation.cs
```diff
@@ -75,6 +22,11 @@ public static class SceneInformation
 
     private static readonly Sprite m_DefaultSprite = Resources.Load<Sprite>("Sprites/default_map");
 
+    private static SCENENAME[] m_ListOfSceneNames = (SCENENAME[])System.Enum.GetValues(typeof(SCENENAME));
+
+    private static SCENENAME[] m_listOfMapNames = m_ListOfSceneNames.Where(map => (map != SCENENAME.MainMenu)).ToArray();
+
```
Gursaanj: Instead of doing the check on the Enum for the main menu or not, I just created two versions where one contains MainMenu and one Doesnt

Assets/Script/Data/SceneInformation.cs
```diff
@@ -164,18 +116,16 @@ public static Sprite PreviewOfScene(MAPNAME map)
     }
 
     /// <summary>
-    /// On the Level Selection Random, Retrieve a random map from the MAPNAME Enum
+    /// On the Level Selection Random, Retrieve a random map from the SCENE Enum
     /// </summary>
-    /// <returns>name of level</returns>
-    private static string RandomizeLevel()
+    /// <returns>name of Map</returns>
+    private static string RandomizeMap()
```
Gursaanj: Needed to change the way levels were being randomized so that Main Menu wasnt called

Assets/Script/Data/SceneInformation.cs
```diff
@@ -60,46 +40,68 @@ public static string MainMenuFileName
     }
 
     /// <summary>
-    /// Gets the list of scene name.
+    /// Gets the list of Map names (i.e not the MainMenu).
     /// </summary>
-    public static SCENENAME[] ListOfSceneNames
+    public static SCENENAME[] ListOfMapNames
```
m4rtint: you can do 
public static SCENENAME[] ListOfMapNames => m_listOfMapNames as shorthand.

Assets/Script/Data/SceneInformation.cs
```diff
@@ -4,49 +4,29 @@
...
 
     private static readonly Sprite m_FountainSprite = Resources.Load<Sprite>("Sprites/fountain");
     private static readonly Sprite m_GazeboSprite = Resources.Load<Sprite>("Sprites/gazebo");
     private static readonly Sprite m_OfficeSprite = Resources.Load<Sprite>("Sprites/office");
 
     private static readonly Sprite m_DefaultSprite = Resources.Load<Sprite>("Sprites/default_map");
 
+    private static SCENENAME[] m_ListOfSceneNames = (SCENENAME[])System.Enum.GetValues(typeof(SCENENAME));
+
+    private static SCENENAME[] m_listOfMapNames = m_ListOfSceneNames.Where(map => (map != SCENENAME.MainMenu)).ToArray();
```
m4rtint: Can you do 
map != SCENENAME.MainMenu  || map != SCENENAME.RANDOM

so that when you do ListOfMapNames in other places,you don't have to keep doing List.Length - 1

Assets/Script/UI/MainMenu/MapSelectionManager.cs
```diff
@@ -88,7 +88,7 @@ private void AnimateMapScale(int currentPointer, bool isLeftPressed, float timeT
 
         // Scale old map smaller
         currentPointer += isLeftPressed ? 1 : -1;
-        currentPointer = Mathf.Clamp(currentPointer, 0, SceneInformation.ListOfSceneNames.Length - 1);
+        currentPointer = Mathf.Clamp(currentPointer, 0, SceneInformation.ListOfMapNames.Length - 1);
```
m4rtint: you can probably take the length from
listOfSceneNames here , so less dependencies on SceneInformation.

Assets/Script/UI/MainMenu/MapSelectionManager.cs
```diff
@@ -61,7 +61,7 @@ private void Update()
     private void ChangeMap(bool isLeftPressed)
     {
         m_pointerToMap += isLeftPressed ? -1 : 1;
-        m_pointerToMap = Mathf.Clamp(m_pointerToMap, 0, SceneInformation.ListOfSceneNames.Length - 1);
+        m_pointerToMap = Mathf.Clamp(m_pointerToMap, 0, SceneInformation.ListOfMapNames.Length - 1);
```
m4rtint: Same here, use listOfSceneNames

Assets/Script/Data/SceneInformation.cs
Gursaanj: smart thinking

Assets/Script/Data/SceneInformation.cs
m4rtint: Why do you need that contains menu?

Assets/Script/UI/MainMenu/MapSelectionManager.cs
Gursaanj: but isnt listofSceneNames a local variable for Initialize()?

Assets/Script/Data/SceneInformation.cs
Gursaanj: actually if we get rid of SCENENAME.RANDOM we wont initialize the random Button right?

